PR #8, opened 2019-03-07
Add link button + fix toggle

dhoko:
Add component for 
- LinkButton

Fix Toggle
- Auto checked value required
- Add mode Icon
- handleChange value
- We can debounce the onChange and change the value from outside

### DIFF
--- a/components/button/LinkButton.js
+++ b/components/button/LinkButton.js
@@ -1,0 +1,20 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import Button from './Button';
+import { getClasses } from '../../helpers/component';
+
+const LinkButton = ({ children, className, ...rest }) => {
+    return (
+        <Button className={getClasses('pm-button-link', className)} {...rest}>
+            {children}
+        </Button>
+    );
+};
+
+LinkButton.propTypes = {
+    children: PropTypes.node,
+    className: PropTypes.string
+};
+
+export default LinkButton;
--- a/components/button/LinkButton.js
+++ b/components/button/LinkButton.js
@@ -2,11 +2,10 @@ import React from 'react';
 import PropTypes from 'prop-types';
 
 import Button from './Button';
-import { getClasses } from '../../helpers/component';
 
 const LinkButton = ({ children, className, ...rest }) => {
     return (
-        <Button className={getClasses('pm-button-link', className)} {...rest}>
+        <Button className={`pm-button-link ${className}`} {...rest}>
             {children}
         </Button>
     );
--- a/components/button/index.js
+++ b/components/button/index.js
@@ -5,3 +5,4 @@ export { default as PrimaryButton } from './PrimaryButton';
 export { default as ButtonGroup } from './ButtonGroup';
 export { default as Group } from './Group';
 export { default as Copy } from './Copy';
+export { default as LinkButton } from './LinkButton';
--- a/components/toggle/Toggle.js
+++ b/components/toggle/Toggle.js
@@ -3,7 +3,6 @@ import PropTypes from 'prop-types';
 import { c } from 'ttag';
 
 import Input from '../input/Input';
-import Label from '../label/Label';
 import Icon from '../icon/Icon';
 
 const label = (type, key) => {
@@ -15,7 +14,7 @@ const label = (type, key) => {
     if (type === 'icon') {
         return (
             <span className="pm-toggle-label-text">
-                <Icon name={key} alt={I18N[key]} />
+                <Icon name={key} alt={I18N[key]} className="pm-toggle-label-img" />
             </span>
         );
     }
@@ -38,10 +37,10 @@ const Toggle = ({ id, checked, onChange, type }) => {
                 className="pm-toggle-checkbox"
                 onChange={() => onChange(value)}
             />
-            <Label htmlFor={id} className="pm-toggle-label" onClick={() => setValue(!value)}>
+            <label htmlFor={id} className="pm-toggle-label" onClick={() => setValue(!value)}>
                 {label(type, 'on')}
                 {label(type, 'off')}
-            </Label>
+            </label>
         </>
     );
 };
--- a/components/toggle/Toggle.js
+++ b/components/toggle/Toggle.js
@@ -1,15 +1,22 @@
-import React from 'react';
+import React, { useState } from 'react';
 import PropTypes from 'prop-types';
 import { t } from 'ttag';
 
 import Input from '../input/Input';
 import Label from '../label/Label';
 
-const Toggle = ({ id, checked }) => {
+const Toggle = ({ id, checked, onChange }) => {
+    const [value, setValue] = useState(checked);
     return (
         <>
-            <Input type="checkbox" id={id} checked={checked} className="pm-toggle-checkbox" />
-            <Label htmlFor={id} className="pm-toggle-label">
+            <Input
+                type="checkbox"
+                id={id}
+                checked={value}
+                className="pm-toggle-checkbox"
+                onChange={() => onChange(value)}
+            />
+            <Label htmlFor={id} className="pm-toggle-label" onClick={() => setValue(!value)}>
                 <span className="pm-toggle-label-text">{t`On`}</span>
                 <span className="pm-toggle-label-text">{t`Off`}</span>
             </Label>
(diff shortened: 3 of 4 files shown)


components/toggle/Toggle.js
```diff
@@ -1,28 +1,60 @@
...
-                <span className="pm-toggle-label-text">{t`Off`}</span>
-            </Label>
+            <Input
+                type="checkbox"
+                id={id}
+                checked={value}
+                className="pm-toggle-checkbox"
+                onChange={() => onChange(value)}
+            />
+            <label htmlFor={id} className="pm-toggle-label" onClick={() => setValue(!value)}>
```
mmso: Should it use `setValue`? Since `checked` is coming from the parent, should it not let the parent also handle the change? E.g. `onChange(!checked)`, so that it can prevent the toggling if it wants?

components/button/LinkButton.js
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import Button from './Button';
+import { getClasses } from '../../helpers/component';
+
+const LinkButton = ({ children, className, ...rest }) => {
+    return (
+        <Button className={getClasses('pm-button-link', className)} {...rest}>
```
EpokK: When we started to write component, we started to use `getClasses` but finally it's not a good idea because the code is super simple as you can see:

```suggestion
        <Button className={`pm-button-link ${className}`} {...rest}>
```

components/button/index.js
```diff
@@ -4,4 +4,5 @@ export { default as LargeButton } from './LargeButton';
 export { default as PrimaryButton } from './PrimaryButton';
 export { default as ButtonGroup } from './ButtonGroup';
 export { default as Group } from './Group';
-export { default as Copy } from './Copy';
\ No newline at end of file
+export { default as Copy } from './Copy';
+export { default as LinkButton } from './LinkButton';
```
EpokK: Can you export `LinkButton` component to the main `index.js` file also?

components/toggle/Toggle.js
```diff
@@ -1,28 +1,60 @@
...
+    id: PropTypes.string.isRequired,
+    type: PropTypes.string,
+    checked: PropTypes.bool.isRequired,
+    onChange: PropTypes.func.isRequired
 };
 
 Toggle.defaultProps = {
-    id: 'toggle'
+    id: 'toggle',
+    type: 'text'
```
EpokK: Give a default `Boolean` value to checked please, like: 

``` js
checked: false
```

components/button/index.js
dhoko: I forgot to add it inside the commit ^^

components/toggle/Toggle.js
```diff
@@ -1,28 +1,60 @@
...
-            <Label htmlFor={id} className="pm-toggle-label">
-                <span className="pm-toggle-label-text">{t`On`}</span>
-                <span className="pm-toggle-label-text">{t`Off`}</span>
-            </Label>
+            <Input
+                type="checkbox"
+                id={id}
+                checked={value}
+                className="pm-toggle-checkbox"
+                onChange={() => onChange(value)}
```
EpokK: Define a `handleChange` function please to `setValue` and call `onChange`.

```suggestion
                onChange={handleChange}
```

To call `onChange` you can also use `useEffect`:

``` js
const handleChange = ({ target }) => {
   setValue(target.checked);
};

useEffect(() => {
   onChange(value);
}, [value]);
```

components/toggle/Toggle.js
```diff
@@ -1,28 +1,61 @@
...
+            />
+            <label htmlFor={id} className="pm-toggle-label" onClick={() => setValue(!value)}>
```
EpokK: Since you use a `label` connected with the input checkbox by the `id`, I don't think you need to define `onClick` attribute here. The `onChange` on `<Input>` is enough.

components/toggle/Toggle.js
```diff
@@ -1,28 +1,61 @@
...
+
+    if (type === 'icon') {
+        return (
+            <span className="pm-toggle-label-text">
+                <Icon name={key} alt={I18N[key]} className="pm-toggle-label-img" />
+            </span>
+        );
+    }
+
+    if (key === 'on') {
```
EpokK: If the design is using `ON` / `OFF` inside a `<Toggle>`, it's a mistake, we always need to use icon.

components/toggle/Toggle.js
```diff
@@ -1,28 +1,61 @@
...
+const Toggle = ({ id, checked, onChange, type }) => {
+    const [value, setValue] = useState(checked);
     return (
         <>
-            <Input type="checkbox" id={id} className="pm-toggle-checkbox" />
-            <Label htmlFor={id} className="pm-toggle-label">
-                <span className="pm-toggle-label-text">{t`On`}</span>
-                <span className="pm-toggle-label-text">{t`Off`}</span>
-            </Label>
+            <Input
```
EpokK: You can use the existing `<Checkbox>` component here.

components/toggle/Toggle.js
dhoko: fixed via useEffect

components/toggle/Toggle.js
dhoko: Before we can use it we need to fix the CSS as the selector won't work.
Ping @nico3333fr the checkbox is made using `+` our Checkbox component is input + span so it won't work.

components/toggle/Toggle.js
```diff
@@ -1,28 +1,62 @@
...
+};
+
+const Toggle = ({ id, checked, onChange }) => {
+    const [value, setValue] = useState(checked);
+
+    /*
+        If you change the checked from the outside -> ex update the prop outside
+        we update the local state with the new value.
+    */
+    useEffect(() => {
```
EpokK: if the `checked` prop change from outside, the component should be re-render, so I don't think this `useEffect` is useful.

components/toggle/Toggle.js
```diff
@@ -1,28 +1,62 @@
...
+        setValue(checked);
+    }, [checked]);
+
+    const handleChange = () => {
+        const newVal = !value;
```
EpokK: It's better to get the value from the `event.target.checked` in case something interact with the `value` in the same component, see https://reactjs.org/docs/forms.html#controlled-components

components/toggle/Toggle.js
```diff
@@ -1,28 +1,62 @@
...
     return (
         <>
-            <Input type="checkbox" id={id} className="pm-toggle-checkbox" />
-            <Label htmlFor={id} className="pm-toggle-label">
-                <span className="pm-toggle-label-text">{t`On`}</span>
-                <span className="pm-toggle-label-text">{t`Off`}</span>
-            </Label>
+            <Input type="checkbox" id={id} checked={value} className="pm-toggle-checkbox" onChange={handleChange} />
+            <label htmlFor={id} className="pm-toggle-label">
+                {label('on')}
```
EpokK: As I said, the toggle can only contains icon, not text.

mmso:
I see the need for aggressively toggling the checkbox, but what do you think about letting `<Toggle/>` just being the basic component with only the `onChange` handler without any `useState` and `useEffect`, and we wrap another component called `AutoToggle` for this purpose:
```
const AutoToggle = ({ onChange, checked, ...rest }) => {
    const [value, setValue] = useState(checked);

    /*
        If you change the checked from the outside -> ex update the prop outside
        we update the local state with the new value.
    */
    useEffect(() => {
        setValue(checked);
    }, [checked]);

    const handleChange = () => {
        const newVal = !value;
        setValue(newVal);
        onChange(newVal);
    };

    return (<Toggle {...rest} onChange={handleChange}/>);
};
```

Because there might be cases where we don't need that behavior

components/toggle/Toggle.js
nico3333fr: Quick detail: there is no more text in the toggle component.

EpokK:
> I see the need for aggressively toggling the checkbox, but what do you think about letting `<Toggle/>` just being the basic component with only the `onChange` handler without any `useState` and `useEffect`, and we wrap another component called `AutoToggle` for this purpose:
> 
> ```
> const AutoToggle = ({ onChange, checked, ...rest }) => {
>     const [value, setValue] = useState(checked);
> 
>     /*
>         If you change the checked from the outside -> ex update the prop outside
>         we update the local state with the new value.
>     */
>     useEffect(() => {
>         setValue(checked);
>     }, [checked]);
> 
>     const handleChange = () => {
>         const newVal = !value;
>         setValue(newVal);
>         onChange(newVal);
>     };
> 
>     return (<Toggle {...rest} onChange={handleChange}/>);
> };
> ```
> Because there might be cases where we don't need that behavior

I prefer if we keep at least the `useState` in this component. I cannot find an example where `Toggle` can be used without a state. See this management as a helper with `handleChange`.

PR #10, opened 2019-03-18
Add layouts

econdepe:
Components for LayoutsSection

### DIFF
--- a/containers/layouts/ComposerModeRadios.js
+++ b/containers/layouts/ComposerModeRadios.js
@@ -1,6 +1,6 @@
 import React from 'react';
 import { c } from 'ttag';
-import { RadioCard, useApiWithoutResult, useMailSettings, useEventManager } from 'react-components';
+import { RadioCards, useApiWithoutResult, useMailSettings, useEventManager } from 'react-components';
 import { updateComposerMode } from 'proton-shared/lib/api/mailSettings';
 import { COMPOSER_MODE } from 'proton-shared/lib/constants';
 import composerPopUpSvg from 'design-system/assets/img/design-system-website/popup.svg';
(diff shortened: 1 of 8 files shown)


containers/layouts/ViewLayoutRadios.js
```diff
@@ -1,24 +1,45 @@
...
+                <img alt="Column" src="null" />
+            </RadioCard>
+            <RadioCard
+                value={ROW}
+                checked={ViewLayout === ROW}
+                id="rowRadio"
+                disabled={loading}
+                name="viewLayout"
+                label={c('Label to change view layout').t`Row`}
+                onChange={handleChange(COLUMN)}
```
EpokK:
```suggestion
                onChange={handleChange(ROW)}
```

containers/layouts/DraftTypeSelect.js
```diff
@@ -1,20 +1,28 @@
...
+    };
 
     const options = [
         { text: c('Option').t`Normal`, value: 'text/html' },
         { text: c('Option').t`Plain Text`, value: 'text/plain' }
     ];
 
-    return <Select options={options} onChange={handleChange} />;
+    return (
+        <>
```
dhoko: Why do you add a fragment here ?

containers/layouts/ViewLayoutRadios.js
```diff
@@ -1,24 +1,45 @@
...
+            <RadioCard
+                value={COLUMN}
+                checked={ViewLayout === COLUMN}
+                id="columnRadio"
+                disabled={loading}
+                name="viewLayout"
+                label={c('Label to change view layout').t`Column`}
+                onChange={handleChange(COLUMN)}
+            >
+                <img alt="Column" src="null" />
```
dhoko: why src=null ?

econdepe: Yeah, Richard already told me. I'm waiting for the images from the design team. I changed null -> todo

containers/layouts/DraftTypeSelect.js
econdepe: I guess the extra <> </> is called a fragment. Actually that is one question I have. When does one have to use it? I used it here because it was used in other LayoutsSection components, but indeed I don't know when to put it

dhoko: You don't need them when you return only one component ex: <Select>.
You need them to prevent an extra div with many elements, ex:

```
<div>
  <ComponentA>....
  <ComponentB>....
</div>
```

vs

```
<>
  <ComponentA>....
  <ComponentB>....
</>
```
cf https://reactjs.org/docs/fragments.html

econdepe: Thanks for the explanation! My bad I haven't reached that section in the React documentation yet

containers/layouts/ComposerModeRadios.js
```diff
@@ -27,7 +27,7 @@ const ComposerModeRadios = () => {
                 label={c('Label to change composer mode').t`Popup`}
                 onChange={handleChange(POPUP)}
             >
-                <img alt="Popup" src="/assets/img/design-system-website/popup.png" />
+                <img alt="Popup" src="assets/img/pm-images/composer-popup.svg" />
```
mmso: Prefer to import it using the regular system

```
import composerPopupSvg from 'assets/img/pm-images/composer-popup.svg';
```
...

later in your component


```
<img alt="Popup" src={composerPopupSvg} />
```

econdepe: Ah, sure! Since I don't know what the folder structure will be eventually (namely what will be the path to fetch the images) I did not bother too much. But let me change it to the React standard to load images.

containers/layouts/ViewLayoutRadios.js
```diff
@@ -1,24 +1,47 @@
...
         <>
-            <Label htmlFor="columnRadio">
-                <Radio id="columnRadio" name="viewLayout" onChange={handleChange(0)} />
-                {c('Label').t`Column`}
-            </Label>
-            <Label htmlFor="rowRadio">
-                <Radio id="rowRadio" name="viewLayout" onChange={handleChange(1)} />
-                {c('Label').t`Radio`}
-            </Label>
+            <RadioCard
```
EpokK: Create a component to manage a list of `RadioCard`.

containers/layouts/ViewModeRadios.js
```diff
@@ -1,24 +1,47 @@
...
+import conversationGroupSvg from 'design-system/assets/img/pm-images/conversation-group.svg';
+import conversationSingleSvg from 'design-system/assets/img/pm-images/conversation-single.svg';
 
-import useApi from '../../hooks/useApi';
+const { GROUP, SINGLE } = VIEW_MODE;
 
 const ViewModeRadios = () => {
-    const api = useApi();
-    const handleChange = (mode) => () => api(updateViewMode(mode));
+    const [{ ViewMode }] = useMailSettings();
```
EpokK: Get `mailSettings` model from parent section via `props`.

containers/layouts/ViewModeRadios.js
```diff
@@ -1,24 +1,47 @@
...
+    const { call } = useEventManager();
```
EpokK: Move all business logic on the parent.

econdepe:
The logic in LayoutsSection has a mistake as @mmso pointed out. I'll try to fix it

PR #13, opened 2019-03-20
Theme section

econdepe:
Closes https://github.com/ProtonMail/protonmail-settings/issues/17


econdepe:
At the moment there no interactivity with the DOM in the ThemeSection. I actually don't know what's are the expected interactions (how many themes should be displayed, whether new themes can be added, what the `preview` and `delete` buttons should do, etc).

Apart from that, there will probably be other things to be corrected

econdepe:
I changed the folder structure based on the folder structure for layouts. The components relevant for themes are now in `react-components/containers/themes`.

econdepe:
The new force-push is to include the color-picker from another branch

econdepe:
Although all the functionality is in, this is still a working version:
- the function `getThemeName` may go into a helpers file.
- the function `getCSS` will be defined in `design-system`.
- maybe we would like to debounce `handleChangeCSSInput`.

Question:
Should the `ThemeCard` for the custom theme be made clickable so that the Modal opens everytime the user clicks on it? Otherwise it's only possible to open it when other theme is selected (because only then the event for change of radio card is fired).

containers/themes/ThemesSection.js
```diff
@@ -0,0 +1,108 @@
...
+        .t`Custom themes from third parties can potentially betray your privacy. Only use themes from trusted sources`;
+
+    return (
+        <>
+            <SubTitle>{c('Title').t`Themes`}</SubTitle>
+            <Alert>{themeText}</Alert>
+            <Alert type="warning">{customizationThemeText}</Alert>
+            <br />
+            <ThemeCards list={availableThemes} themeName={themeName} onChange={handleChangeTheme} disabled={loading} />
+            <ConfirmModal
```
EpokK: Extend the `InputModal` to display an `<input />` or a `<textarea>` please.

containers/themes/ThemeCard.js
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
```
EpokK: I don't see the value of creating this component instead of directly using `RadioCard`.

containers/themes/ThemesSection.js
```diff
@@ -0,0 +1,108 @@
...
+    TextArea
+} from 'react-components';
+import { updateTheme } from 'proton-shared/lib/api/mailSettings';
+import { themeDark, themeLight, themeBlue, themeCustom } from './availableThemes.js';
+
+const availableThemes = [themeDark, themeLight, themeBlue, themeCustom];
+
+const getThemeName = (Theme) => {
+    if (Theme) {
+        if (/\/* light-theme \/*/.test(Theme)) {
```
EpokK: With `startsWith` instead of the RegExp please.

containers/themes/ThemesSection.js
```diff
@@ -0,0 +1,108 @@
...
+        setCustomCSSInput('');
+        await request(getCSS(themeName));
+        call();
+    };
+    const handleConfirm = async () => {
+        setShowCustomModal(!showCustomModal);
+        await request(customCSSInput);
+        call();
+    };
+    const handleClose = () => {
```
EpokK: One line please
```js
    const handleClose = () => setShowCustomModal(false);
```

containers/themes/ThemesSection.js
```diff
@@ -0,0 +1,108 @@
...
+        setShowCustomModal(false);
+    };
+    const handleChangeCSSInput = (event) => {
```
EpokK: Do it in one line also:
``` js
const handleChangeCSSInput = ({ target }) => setCustomCSSInput(target.value);
```

containers/themes/ThemesSection.js
```diff
@@ -0,0 +1,108 @@
...
+            return CSSLightTheme;
+        case 'blue-theme':
+            return CSSBlueTheme;
+        default:
+            return '';
+    }
+};
+
+const ThemesSection = () => {
+    const [showCustomModal, setShowCustomModal] = useState(false);
```
EpokK: Use `useModal` hook for that.

containers/themes/ThemesSection.js
```diff
@@ -0,0 +1,108 @@
...
+
+const CSSLightTheme = '/* light-theme */';
```
EpokK: Extract these constants in `proton-shared/lib/constants` please.

containers/themes/availableThemes.js
```diff
@@ -0,0 +1,36 @@
+import themeDarkSvg from 'design-system/assets/img/pm-images/theme-dark.svg';
+import themeLightSvg from 'design-system/assets/img/pm-images/theme-light.svg';
+import themeBlueSvg from 'design-system/assets/img/pm-images/theme-blue.svg';
+import themeTestSvg from 'design-system/assets/img/pm-images/theme-test.svg';
+
+export const themeDark = {
+    label: 'Dark (Default)',
+    id: 'dark',
+    value: 'dark-theme',
```
EpokK: Use constants from `proton-shared/lib/constants` here please.

containers/themes/ThemesSection.js
```diff
@@ -0,0 +1,108 @@
...
+    SubTitle,
+    Alert,
+    ThemeCards,
+    ConfirmModal,
+    TextArea
+} from 'react-components';
+import { updateTheme } from 'proton-shared/lib/api/mailSettings';
+import { themeDark, themeLight, themeBlue, themeCustom } from './availableThemes.js';
+
+const availableThemes = [themeDark, themeLight, themeBlue, themeCustom];
```
EpokK: You should get the list from `proton-shared/lib/constants`:
```js
export const LIGHT_THEME = '/* light-theme */';
export const DARK_THEME = '/* dark-theme */';
export const BLUE_THEME = '/* blue-theme */';
export const THEMES = [LIGHT_THEME, DARK_THEME, BLUE_THEME];
```

containers/themes/ThemesSection.js
econdepe: I did some modification to `InputModal`. In particular I added a `useEffect` so that the component would update if it's closed (otherwise it retains the input from the last time it was open, even if clicking `cancel`).

However, to match the design https://user-images.githubusercontent.com/33841139/56794123-caee7700-680d-11e9-9585-edda1b962c71.png the component should be further modified (to fit the `Alert`). I wonder if it's better to create a new component `TextAreaModal`.

containers/themes/ThemeCard.js
econdepe: Yes, originally `ThemeCard` was a bit different, that's why it was a different component. I added a `customizable` prop now, so I think it makes sense to have the component again.

containers/themes/ThemesSection.js
econdepe: Done!

containers/themes/ThemesSection.js
econdepe: Done!

containers/themes/ThemesSection.js
econdepe: Easy piecey!

containers/themes/ThemesSection.js
econdepe: Yep, the code looks way less cluttered now

containers/themes/ThemesSection.js
econdepe: These are just temporary. I'm waiting for Nico to provide with the CSS. These constants will never be in the code, actually. The CSS will be generated from a function in `design-system`.

containers/themes/ThemesSection.js
econdepe: I followed a slightly different road with the last commit (0285b25)

econdepe:
Current status:

- Waiting for feedback on the use of constants I proposed.
- There's some code commented at the end of `ThemeSection.js`. Waiting for the themes CSS to be produced.
- There's some code commented in the middle of `ThemeSection.js`. The `InputModal` component is not flexible enough to account for the desired design. Either that component is considerably modified, or a new component, `TextAreaModal` is created. Waiting for feedback on which road to follow (since I'm not sure in what other places these components are used).
- Should we set some conditions on the length of the text users can enter in the Custom CSS area?

helpers/themes.js
```diff
@@ -0,0 +1,26 @@
...
+ */
+export const getThemeName = (Theme) => {
+    if (Theme) {
+        if (Theme.startsWith(LIGHT_THEME_IDENTIFIER)) {
+            return LIGHT_THEME_VALUE;
+        }
+        if (Theme.startsWith(BLUE_THEME_IDENTIFIER)) {
+            return BLUE_THEME_VALUE;
+        }
+        return CUSTOM_THEME_VALUE;
```
EpokK:
```suggestion
        return Theme;
```

helpers/themes.js
```diff
@@ -0,0 +1,26 @@
...
+    }
+    return DARK_THEME_VALUE;
```
EpokK:
```suggestion
    return '';
```

helpers/themes.js
econdepe: see new proposal

helpers/themes.js
econdepe: see new proposal

econdepe:
# Current status:

The current version of `ThemesSection` requires an input component that has the props:
- `type`, which can be `'textarea'`.
- `subtitle`, in order to fit the design in https://github.com/ProtonMail/protonmail-settings/issues/17 . The modal there has a title section ('Custom Theme'), and another section with some text alert. the `subtitle` prop would take care of the text alert.

# Question
Should the image for the custom theme be dynamic? Otherwise the user can set a custom theme that looks nothing like the picture for the custom theme. This is not great, but I'm not sure how easy/dangerous would be have a dynamic image for the theme. Another possibility is to have an image for the custom theme that makes more clear it's a custom one.

components/modal/Input.js
```diff
@@ -17,6 +17,7 @@ Good candidates:
 */
 
 const InputModal = ({
+    isTextArea,
```
EpokK: Use `type` (String), much more flexible.

components/modal/Input.js
```diff
@@ -75,6 +92,7 @@ InputModal.propTypes = {
 };
 
 InputModal.defaultProps = {
+    isTextArea: false,
```
EpokK:
```suggestion
    type: 'input',
```

components/modal/Input.js
econdepe: done

components/modal/Input.js
econdepe: done

PR #16, opened 2019-03-22
[Button] - Allow us to extend classNames

dhoko:
Today: 
`<Button className="mlauto">-></Button>` = :boom: no more styles :/

With the fix:
Screenshot from 2019-03-22 12-31-52

It works.
btw we should not have mr1 inside the button imho

### DIFF
--- a/components/button/Button.js
+++ b/components/button/Button.js
@@ -62,7 +62,7 @@ const Button = (props) => {
         <button
             role={role}
             disabled={loading ? true : disabled}
-            className={className}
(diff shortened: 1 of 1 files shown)


components/button/Button.js
```diff
@@ -98,7 +98,7 @@ Button.propTypes = {
 };
 
 Button.defaultProps = {
-    className: 'pm-button mr1',
+    className: '',
```
EpokK: Remove this default attribute then.

PR #17, opened 2019-03-22
fix route for images

econdepe:


EpokK:
Wrong path, I already fixed the issue.

PR #29, opened 2019-04-01
Keep state for the modal

dhoko:
use case, remove a filter post confirmation
```js

const handleClickDelete = (filter) => () => {
    openConfirmModal(filter);
};

const handleConfirmConfirmModal = async () => {
    await removeFilter(stateConfirmModal.data);
    closeConfirmModal();
}
```


EpokK:
As discussed with @dhoko this is not necessary.

PR #32, opened 2019-04-04
Observer Section

econdepe:
The code below is not working. Although the observers in `ObserverSection` are working properly (you can monitor them through console.logs), I'm trying to pass the intersectionRatios they produce to a parent component `ObserverSections`, but that step is failing.
In the parent I created an array with `useState` that should store the values of intersectionRatios produced by the children components. However instead of updating, this array seems to reset to the original value everytime the observers emit an event, which is not the intended behavior.

Note: Solving the above problem will not produce the exact desired behavior. In order to do a more faithful comparison of intersection ratios I think we need the observers to trigger events not just when crossing the 50% threshold, but for an array of thresholds like [0, 0.1, 0.2, ..., 0.9, 1]. But in principle this can be easily done later after fixing the above problem.


econdepe:
Don't pay attention to the line
```
import { ObserverSection } from 'react-components';
```
in `ObserverSections.js`. I was running tests inside an environment were ObserverSection was properly exported.

econdepe:
The main problem I described at the beginning is solved by using [functional updates](https://reactjs.org/docs/hooks-reference.html#functional-updates). Now the ObserverSections does what intended to. But I still need to implement the array of thresholds (instead of a single threshold) so that the behavior is exactly what we want.

econdepe:
The two problems mentioned at the very beginning are solved, and the ObserverSection parent and child components are doing what they're supposed to.

econdepe:
Comments on the last commits:

1. As pointed out by @dhoko , I was merging instead of rebasing. Hopefully this has been fixed by the last force push.

2. There seems to be some issue with webpack and IntersectionObserver. It seems webpack does not recognize it. The recommended solution I found on the internet (https://www.robinwieruch.de/react-intersection-observer-api/) was to load the polyfill package `intersection-observer` (https://github.com/w3c/IntersectionObserver). I had to require it by hand in `IntersectionObserver.js`.

3. For the dynamic changing (as the user scrolls up/down) of the URL hash, I found that one needs to use `history.replace` (https://developer.mozilla.org/en-US/docs/Web/API/History_API). The `window.location` methods reload the page which is not the behavior we want. However, opposite to the these methods, `history.replace` does not fire a `hashchange` event, which the component `SubSidebar` is expecting. In the code I'm firing the `hashchange` by hand, and I checked that `SubSidebar` receives these events properly, but this may be a bit ugly code-wise. The other solution would be to modify the code of `SubSidebar`.

4. In the localhost tests I was running, there was an issue with relative URLs. Whenever I use features that require relative URLs, for instance an `href=#...`, the browser never recognizes that it should point to a relative URL, but it rather points to `localhost/#...`. I guess this is probably a problem of my testing environment, but because of that I commented line 20 of `ObserverSections.js` and I'm using line 19 instead. I think line 20 should work, but I can't check it in my testing environment.

components/sidebar/ObserverSection.js
```diff
@@ -0,0 +1,71 @@
+import React, { useEffect } from 'react';
```
EpokK: Put this component and others under `components/container` please.

components/sidebar/ObserverSection.js
```diff
@@ -0,0 +1,71 @@
...
+    rootMargin: PropTypes.string,
+    granularity: PropTypes.number.isRequired,
+    index: PropTypes.number.isRequired,
+    setIntersectionRatios: PropTypes.func.isRequired,
+    children: PropTypes.node.isRequired
+};
+
+ObserverSection.defaultProps = {
+    rootElement: null,
+    rootMargin: '0px'
```
EpokK: Set a default value for `granularity` please.

components/sidebar/ObserverSections.js
```diff
@@ -0,0 +1,56 @@
...
+
+        // replace URL
+        const currentURL = document.URL;
+        const newURL = /#/.test(currentURL) ? currentURL.replace(/#(.*)/, `#${idOfMax}`) : currentURL + `#${idOfMax}`;
+        // const newURL =  `#${idOfMax}`;   // This should work instead of the previous two lines, but local testing seems to miss relative URLs
+        if (newURL !== currentURL) {
+            history.replaceState({ section: `${idOfMax}` }, '', newURL);
+            window.dispatchEvent(new HashChangeEvent('hashchange'));
+        }
+    });
```
EpokK: If you don't set a second parameter for `useEffect`, it will run his callback each time the app refresh. Could you set a dependency?

components/sidebar/ObserverSections.js
```diff
@@ -0,0 +1,56 @@
...
+const ObserverSections = ({ list }) => {
+    const [intersectionRatios, setIntersectionRatios] = useState(
+        Array(list.length)
+            .fill(1)
+            .fill(0, 1)
+    );
+
+    useEffect(() => {
+        // find id corresponding to biggest intersection ratio
+        const indexOfMax = (arr) => arr.findIndex((elm) => elm === Math.max(...arr));
```
EpokK: Move `indexOfMax` outside of the `useEffect`. Maybe even outside this file inside a helper file.

components/sidebar/ObserverSections.test.js
```diff
@@ -0,0 +1,35 @@
...
+describe('ObserverSections component', () => {
+    const dummyText =
+        'Lorem ipsum dolor sit amet, consectetur adipiscing elit. Sed pretium enim nec massa fringilla, ac ultrices tortor posuere. Fusce sed quam vitae arcu pharetra congue. Quisque in elementum nibh.';
+    const repeatText = (text, n) => {
+        return Array(n)
+            .fill(text)
+            .reduce((acc, cur) => cur + acc, '');
+    };
+    const list = [
+        { id: 'section1', granularity: 20, children: <div id="text-1">{repeatText(dummyText, 1)}</div> },
```
EpokK: Why do you need to set `id="text-1"`?

components/sidebar/ObserverSections.js
econdepe: What should it depend on?

components/sidebar/ObserverSections.test.js
econdepe: It's not needed for anything. I will remove it.

components/sidebar/ObserverSection.js
```diff
@@ -0,0 +1,71 @@
...
+        const steps = parseInt(granularity, 10);
+        const thresholds = [];
+
+        for (let i = 0; i <= steps; i++) {
+            thresholds.push(i / steps);
+        }
+        return thresholds;
+    };
+
+    const handleIntersect = (entries) => {
```
EpokK: Could we debounce this function to 500ms to avoid overcharging on his parent?

components/sidebar/ObserverSection.js
```diff
@@ -0,0 +1,71 @@
...
+    // the granularity prop should be an integer and is meant to mark how many divisions we make to
+    // an area in order to observe it with an intersectionObserver.
+    //
+    // E.g. granularity = 2 will mean that intersectionObserver will emit events when
+    // the 0%, 50% or 100% of the area overlaps with a given target area.
+    //
+    // In general, the higher the granularity, the more sensitive we will be to changes in overlapping areas.
+    // Probably granularity = 20 will suffice for the purpose of updating URL in function of the dislpay area
+
+    const buildThresholds = (granularity) => {
```
EpokK: Extract this function to helpers please.

components/sidebar/ObserverSection.js
```diff
@@ -0,0 +1,71 @@
...
+        for (let i = 0; i <= steps; i++) {
+            thresholds.push(i / steps);
+        }
+        return thresholds;
+    };
+
+    const handleIntersect = (entries) => {
+        entries.forEach(function(entry) {
+            setIntersectionRatios((intersectionRatios) => {
+                let newIntersectionRatios = intersectionRatios.slice();
```
EpokK: Use `const`
```suggestion
                const newIntersectionRatios = intersectionRatios.slice();
```

components/sidebar/ObserverSection.js
econdepe: done!

components/sidebar/ObserverSections.js
econdepe: Done!

components/sidebar/ObserverSections.js
econdepe: Done both!

components/sidebar/ObserverSection.js
econdepe: Done!

components/sidebar/ObserverSection.js
econdepe: Done both!

components/sidebar/ObserverSection.js
econdepe: I was misled by the fact that I was modifying the array. But indeed the pointer does not change. Done!

components/sidebar/ObserverSection.js
```diff
@@ -0,0 +1,71 @@
...
+                newIntersectionRatios[index] = Math.min(entry.intersectionRatio, 1); // manual fix for bug IntersectionObserverEntry.intersectionRatio > 1
+                return newIntersectionRatios;
```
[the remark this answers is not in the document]
econdepe: The return is required by the way functional updates work: https://reactjs.org/docs/hooks-reference.html#functional-updates
If I don't include the return, the application crashes

components/sidebar/ObserverSection.js
```diff
@@ -0,0 +1,71 @@
...
+        const thresholds = [];
+
+        for (let i = 0; i <= steps; i++) {
+            thresholds.push(i / steps);
+        }
+        return thresholds;
+    };
+
+    const handleIntersect = (entries) => {
+        entries.forEach(function(entry) {
```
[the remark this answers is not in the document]
econdepe: I copied and pasted from somewhere else. Done!

components/sidebar/ObserverSection.js
```diff
@@ -0,0 +1,71 @@
...
+                return newIntersectionRatios;
+            });
+        });
+    };
+
+    useEffect(() => {
+        // Create an observer that triggers when intersectionRatio changes 1/granularity
+        const options = {
+            root: rootElement,
+            rootMargin: rootMargin,
```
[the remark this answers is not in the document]
econdepe: True. Done!

components/sidebar/ObserverSections.test.js
```diff
@@ -0,0 +1,35 @@
...
+        { id: 'section2', granularity: 20, children: <div id="text-2">{repeatText(dummyText, 2)}</div> },
+        { id: 'section3', granularity: 20, children: <div id="text-3">{repeatText(dummyText, 3)}</div> },
+        { id: 'section4', granularity: 20, children: <div id="text-4">{repeatText(dummyText, 4)}</div> },
+        { id: 'section5', granularity: 20, children: <div id="text-5">{repeatText(dummyText, 5)}</div> }
+    ];
+
+    it('should render the observer sections properly', () => {
```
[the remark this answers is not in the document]
econdepe: Yeah, the testing was at a preliminary stage. Although the main testing is really whether the `ObserverSections` respond properly to the user scrolling. I checked this in my testing environment, but don't know if it's possible to do such testing with mock functions.

econdepe:
The CircleCI fails because of importing in `ObseverSection.js` through a local route:
```
import { debounce } from '../../../proton-shared/lib/helpers/function';
```
`npm test` is not giving errors in my local machine

components/sidebar/ObserverSections.js
```diff
@@ -0,0 +1,56 @@
...
+        const idOfMax = list[indexOfMax(intersectionRatios)].id;
```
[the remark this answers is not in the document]
econdepe: Done!

PR #34, opened 2019-04-05
Easier to extend classNames for the Icon

dhoko:
`<Icon className="ml1" />` didn't work.

Screenshot from 2019-04-05 14-27-31

New props:
- `fill`: default value **grey** -> fill color for the icon
- `size`: default value **16** -> size of the icon

### DIFF
--- a/components/icon/Icon.js
+++ b/components/icon/Icon.js
@@ -13,7 +13,7 @@ import PropTypes from 'prop-types';
 const Icon = ({ name, className, viewBox, alt, ...rest }) => {
     return (
         <>
-            <svg viewBox={viewBox} className={className} role="img" focusable="false" {...rest}>
+            <svg viewBox={viewBox} className={`icon-16p `.concat(className)} role="img" focusable="false" {...rest}>
                 <use xlinkHref={`#shape-${name}`} />
             </svg>
             {alt ? <span className="sr-only">{alt}</span> : null}
@@ -30,7 +30,7 @@ Icon.propTypes = {
 
 Icon.defaultProps = {
     viewBox: '0 0 16 16',
-    className: 'icon-16p fill-global-grey'
+    className: ''
 };
 
 export default Icon;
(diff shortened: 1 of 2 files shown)


components/icon/Icon.js
```diff
@@ -13,7 +13,7 @@ import PropTypes from 'prop-types';
 const Icon = ({ name, className, viewBox, alt, ...rest }) => {
     return (
         <>
-            <svg viewBox={viewBox} className={className} role="img" focusable="false" {...rest}>
+            <svg viewBox={viewBox} className={`icon-16p `.concat(className)} role="img" focusable="false" {...rest}>
```
EpokK: `icon-16p` was put inside `className` as default value to make it overridable, example:

```jsx
<Icon className="icon-32p fill-global-grey mr1" />
```

components/icon/Icon.js
```diff
@@ -30,7 +30,7 @@ Icon.propTypes = {
 
 Icon.defaultProps = {
     viewBox: '0 0 16 16',
-    className: 'icon-16p fill-global-grey'
```
EpokK: Are you sure `fill-global-grey` is not anymore needed?

components/icon/Icon.js
dhoko: yep but the size should be an arg as the name imho. Extend classnames is more common than changing the size

components/icon/Icon.js
dhoko: cf screenshot

components/icon/Icon.js
dhoko: both color and size are config, we should change them via attribute as we do for the name, I will add this

components/icon/Icon.js
nico3333fr: TBH, I'm not sure this will work on all devices (should be more tested, I don't know how browsers will behave in this case).
By default, this is black, not using Proton's global grey.

dhoko: `Dans l'absolu, on ne devrait jamais l'instancier sans une couleur ces icônes, sauf en cas de composant particulier comme le toggle`

So default grey and for Toggle we use none

PR #36, opened 2019-04-11
Add bug report modal

EpokK:
### Added

* useConfig (React hook)
* ConfigContext
* BugModal
* FileInput

### Screenshots

Capture d’écran 2019-04-12 à 08 34 14
Capture d’écran 2019-04-12 à 08 34 24

### Note

- We hide by default system information (browser, os, etc...)
- We could add a `<select>` to set a title for the issue (customer support needs to provide the list of options)


containers/support/AttachScreenshot.js
```diff
@@ -0,0 +1,63 @@
...
+                await Promise.all(
+                    images.map((img) =>
+                        resize(img, MAX_SIZE_SCREENSHOT).then((base64str) => ({
+                            name: img.name,
+                            blob: toBlob(base64str)
+                        }))
+                    )
+                )
+            );
+        } else {
```
dhoko: use a return first ;) 
if no images -> error

containers/support/AttachScreenshot.js
```diff
@@ -0,0 +1,63 @@
...
+                text: c('Error notification in the bug report modal when the user upload file').t`No image selected`
+            });
+        }
+    };
+
+    if (attached) {
+        return (
+            <>
+                <Text>
+                    <Icon name="insert-image" /> {c('Info').t`Screenshot(s) attached`}
```
dhoko: Maybe we can display a small preview (~ 500*400) ?

containers/support/BugModal.js
```diff
@@ -0,0 +1,185 @@
...
+    const os = getOS();
+    const browser = getBrowser();
+    const device = getDevice();
+    const [model, update] = useState({
+        OS: os.name,
+        OSVersion: os.version || '',
+        Browser: browser.name,
+        BrowserVersion: browser.version,
+        Resolution: `${window.innerHeight} x ${window.innerWidth}`,
+        Client: 'Angular', // TODO define a new one with the API
```
dhoko: We should pass the config from the app as a prop to the modal, so we can extend it with specific keys

containers/support/BugModal.js
```diff
@@ -0,0 +1,185 @@
...
+        ClientVersion: '3.15.23', // TODO get it from app
+        ClientType: 1, // TODO get it from config
+        DeviceName: device.vendor,
+        DeviceModel: device.model,
+        Title: `[ProtonMail Settings React] Bug [${location.path}]`,
```
dhoko: Idem for the title, it should comes as a prop --> `ProtonMail Settings React`

containers/support/BugModal.js
```diff
@@ -0,0 +1,185 @@
...
+    const { request, loading } = useApiWithoutResult(reportBug);
+    const link = (
+        <Href url="https://protonmail.com/support/knowledge-base/how-to-clean-cache-and-cookies/">{c('Link')
+            .t`clearing your browser cache`}</Href>
+    );
+    const handleChange = (key) => ({ target }) => update({ ...model, [key]: target.value });
+
+    const handleReset = () => {
+        // eslint-disable-next-line no-unused-vars
+        const { Attachments, ...newModel } = model;
```
dhoko: Can you create a omit function or import the one from lodash ?

containers/support/BugModal.js
```diff
@@ -0,0 +1,185 @@
...
+        update(newModel);
+    };
+
+    const handleUpload = (images) => update({ ...model, Attachments: images });
+
+    const format = (parameters = {}) => {
```
dhoko: doc ?

containers/support/BugModal.js
```diff
@@ -0,0 +1,185 @@
...
+        return Object.entries(parameters).reduce((acc, [key, value]) => {
```
dhoko: Object.values().reduce(acc, key) => {
const value = parameters[key
}

containers/support/BugModal.js
```diff
@@ -0,0 +1,185 @@
...
+            if (value instanceof Array) {
```
dhoko: Why not Array.isArray ? Do we have a specific use case ?

containers/support/AttachScreenshot.js
EpokK: I let design team design.

containers/support/BugModal.js
EpokK: A `Context` is probably more appropriate.

containers/support/BugModal.js
EpokK: You are right.

containers/support/BugModal.js
EpokK: Good idea.

containers/support/BugModal.js
EpokK: What's wrong with `Object.entries`?

containers/support/BugModal.js
EpokK: I will change it, `instanceof` was previously used in the part of code I copied.

containers/support/BugModal.js
```diff
@@ -0,0 +1,183 @@
...
+        <Href url="https://protonmail.com/support/knowledge-base/how-to-clean-cache-and-cookies/">{c('Link')
+            .t`clearing your browser cache`}</Href>
+    );
+    const handleChange = (key) => ({ target }) => update({ ...model, [key]: target.value });
+
+    const getParameters = () => {
+        return images.reduce((acc, { name, blob }) => {
+            acc[name] = blob;
+            return acc;
+        }, model);
```
dhoko: :warning: side effect

containers/support/BugModal.js
```diff
@@ -0,0 +1,183 @@
...
+    const { CLIENT_ID, APP_VERSION } = useConfig();
+    const Client = CONFIGS[CLIENT_ID];
+    const [{ Name = '' }] = useUser();
+    const { createNotification } = useNotifications();
+    const [addresses = []] = useAddresses();
+    const [{ Email = '' } = {}] = addresses;
+    const os = getOS();
+    const browser = getBrowser();
+    const device = getDevice();
+    const [model, update] = useState({
```
dhoko: imho the defaut object should come from a helper, not as a default inside the modal

containers/support/BugModal.js
```diff
@@ -0,0 +1,183 @@
...
+    PrimaryButton,
+    useApiWithoutResult,
+    useUser,
+    useAddresses,
+    useNotifications,
+    useConfig
+} from 'react-components';
+import AttachScreenshot from './AttachScreenshot';
+
+const CONFIGS = {
```
dhoko: idem should be inside a helper

containers/support/AttachScreenshot.js
```diff
@@ -0,0 +1,63 @@
...
+        if (!images.length) {
+            return createNotification({
+                type: 'error',
+                text: c('Error notification in the bug report modal when the user upload file').t`No image selected`
+            });
+        }
+
+        setAttached(true);
+        onUpload(
+            await Promise.all(
```
dhoko: it should be a helper.

containers/support/AttachScreenshot.js
```diff
@@ -0,0 +1,63 @@
...
+    const [attached, setAttached] = useState(false);
+    const { createNotification } = useNotifications();
+
+    const handleClick = () => {
+        setAttached(false);
+        onReset();
+    };
+
+    const handleChange = async ({ target }) => {
+        const images = [...target.files].filter(({ type }) => /^image\//i.test(type));
```
dhoko: idem should be a hlper

containers/support/AttachScreenshot.js
EpokK: I didn't get this one. It's a `Promise.all` running a helper function and formatting a specific result. Also, the logic is already extracted inside a helper.

containers/support/AttachScreenshot.js
EpokK: This is a filter on a group of file to extract images. That's specific to this upload I think.

PR #44, opened 2019-04-17
Color Picker for issue #39

### LINKED ISSUE #39: Add color picker
https://github.com/Simonwep/pickr can be a good candidate.

econdepe:
After much fight with the `color-react` library!

Closes #39


components/input/ColorPicker.js
```diff
@@ -0,0 +1,96 @@
...
+        display: false,
+        color: initialColor
+    });
+
+    const rgbaColor = (color) => {
+        return `rgba(${color.rgb.r}, ${color.rgb.g}, ${color.rgb.b}, ${color.rgb.a})`;
+    };
+
+    const handleClick = () => {
+        setPickerState((prevState) => {
```
EpokK: Shorter version:
``` js
setPickerState({
   ...pickerState,
   display: !pickerState.display
});
```

components/input/ColorPicker.js
```diff
@@ -0,0 +1,96 @@
...
+            return {
+                ...prevState,
+                display: !pickerState.display
+            };
+        });
+    };
+
+    const handleClose = () => {
+        setPickerState((prevState) => {
```
EpokK: You can use a shorter version also here.

components/input/ColorPicker.js
```diff
@@ -0,0 +1,96 @@
...
+            return {
+                ...prevState,
+                display: false
+            };
+        });
+    };
+
+    const handleChange = (color) => {
+        setPickerState((prevState) => {
```
EpokK: Same: shorter version please.

components/input/ColorPicker.js
```diff
@@ -0,0 +1,96 @@
...
+            return {
+                ...prevState,
+                color: color
+            };
+        });
+    };
+
+    const popover = {
```
EpokK: Can you define CSS in a separate file and import it please? Add a `className` to the component if needed.

components/input/ColorPicker.js
```diff
@@ -0,0 +1,96 @@
...
+    );
+
+    useEffect(() => {
+        if (onColorChange) {
+            onColorChange();
+        }
+    }, [pickerState.color]);
+
+    return (
+        <div style={{ position: 'relative' }}>
```
EpokK:
```suggestion
        <div className="relative">
```

components/input/ColorPicker.js
econdepe: The shorter version that you propose does work, but I'm a bit puzzled. Since the new state is making use of the previous state, I thought one would need a [functional update](https://reactjs.org/docs/hooks-reference.html#functional-updates). That's why I fed a function to `setPickerState`... :thinking:

econdepe: I decided to split the state object into two variables since they're not really correlated. I followed the recommendations from the [React docs](https://reactjs.org/docs/hooks-faq.html#should-i-use-one-or-many-state-variables).
The code is cleaner and as a byproduct this issue does not exist (at any rate the recommendation by the React guide is to use the functional update as I was doing)

components/input/ColorPicker.js
econdepe: Done!

components/input/ColorPicker.js
econdepe: See comment above

components/input/ColorPicker.js
econdepe: See comment above

components/input/ColorPicker.js
econdepe: I moved the definitions to some auxiliary CSS, but I don't know what are the good practices for CSS or where to put this auxiliary file. Probably some changes to what I did are required.

As designed at present, if a `className` is specified, it would go under `...rest` and passed down to the button. Do you think a different functionality is needed?

EpokK: As you did is OK for me. Just rename the file to `ColorPicker.css` to be consistent with the component name ;)

components/input/ColorPicker.css
```diff
@@ -0,0 +1,11 @@
+.popover {
+    position: absolute;
+    z-index: 2;
+}
+.cover {
+    position: fixed;
+    top: 0px;
```
nico3333fr: If the value is `0`, there is no need to have a unit ;)

econdepe:
How the color picker looks like without text:
color-picker-default

econdepe:
It's possible to have text in the button for the color picker, although at present the text color is always black, which is not great. However at the moment there is nowhere where a color picker with text is needed.
color-picker-with-text
It would be easy to modify this feature later.

components/input/ColorPicker.css
econdepe: I suspected that. Let me change it!

PR #47, opened 2019-04-24
WIP: create color picker with input

### LINKED ISSUE #46: Create color picker with input
In order to match the presumable UX of the design in https://app.zeplin.io/project/5bb1deae290afc111d140a3f/screen/5c4eb40e54a4b17c6ed5af5a , I think we need a new component that manages together the color picker and the input field, so that one is responsive to the other (meaning a change of color must be reflected in the hex string displayed in the input field and a change in the input field must be reflected in the color picker).

econdepe:
Closes #46

### DIFF
--- a/components/input/ColorPicker.js
+++ b/components/input/ColorPicker.js
@@ -4,7 +4,7 @@ import { Button } from 'react-components';
 import { ChromePicker } from 'react-color';
 import './ColorPicker.css';
 
-const ColorPicker = ({ text, initialRgbaColor, onColorChange, ...rest }) => {
+const ColorPicker = ({ text, initialRgbaColor, onChange, ...rest }) => {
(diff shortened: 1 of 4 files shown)


econdepe:
I don't see how to simply reuse the color picker component as child (at least without twisting too much the code, I think), because the "selected color" prop never leaves that child component, and it would need to be passed to the parent component. Therefore I re-used much of the code of `ColorPicker`.

econdepe:
Thanks to a suggestion by @dhoko , the `ColorPickerWithInput` component is now using the `ColorPicker` one. For that, I had to change slightly the `ColorPicker` code, and I think both components are looking better now while keeping the proper functionality.

Left to do: Make the button look better

econdepe:
Closed after issue #46 has been closed (due to changes in the design)

PR #49, opened 2019-04-24
Improvements to ColorPicker

econdepe:
Two new changes:
- I changed the prop name `onColorChange` to `onChange` since that is the common notation everywhere else. Also this prop is now passed down to `ChromePicker` so that we can control what happens with color changes from parent components.
- The `useEffect` hook is triggered by changes in the prop `initialRgbaColor`. In this way we can also change the color of the color picker from parent components.


econdepe:
sorry, I was tricked by github into merging. I will fix it.

PR #50, opened 2019-04-24
Improvements to Color Picker

econdepe:
Two new changes:
- I changed the prop name onColorChange to onChange since that is the common notation everywhere else. Also this prop is now passed down to ChromePicker so that we can control what happens with color changes from parent components.
- The useEffect hook is triggered by changes in the prop initialRgbaColor. In this way we can also change the color of the color picker from parent components.

### DIFF
--- a/components/input/ColorPicker.js
+++ b/components/input/ColorPicker.js
@@ -1,5 +1,6 @@
 import React, { useState, useEffect } from 'react';
 import PropTypes from 'prop-types';
+import { noop } from 'proton-shared/lib/helpers/function';
 import { Button } from 'react-components';
 import { ChromePicker } from 'react-color';
 import './ColorPicker.scss';
@@ -19,9 +20,7 @@ const ColorPicker = ({ text, initialRgbaColor, onChange, ...rest }) => {
     };
     const handleChange = (color) => {
         setRgbaColor(color.rgb);
-        if (onChange) {
-            onChange(color);
-        }
+        onChange(color);
     };
 
     useEffect(() => {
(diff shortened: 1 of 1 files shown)


components/input/ColorPicker.js
```diff
@@ -19,21 +19,22 @@ const ColorPicker = ({ text, initialRgbaColor, onColorChange, ...rest }) => {
     };
     const handleChange = (color) => {
         setRgbaColor(color.rgb);
+        if (onChange) {
```
dhoko: You can use a noop as default if you want via proptypes. (noop is inside proton-shared/funcitons)

components/input/ColorPicker.js
```diff
@@ -19,21 +19,22 @@ const ColorPicker = ({ text, initialRgbaColor, onColorChange, ...rest }) => {
...
+            onChange(color);
+        }
     };
 
+    useEffect(() => {
```
EpokK: if it's the "initial" value, you don't need to use `useEffect` here. I don't expect to receive a change from `initialRgbaColor` right?

econdepe: Yeah, for the component as it is it's enough. But when I was building the `ColorPickerWithInput`, if I wanted to use the `ColorPicker` component there, I needed a way for the `Input` component to tell the color chosen in the input field to `ColorPicker`, and the natural way to do it seemed to me through `initialRgbaColor`. That's why I wrote the `useEffect` hook like that. For the current functoinality is not needed, but if we later need a way to interact externally with the color displayed in the color picker, I think doing it through `initialRgbaColor` would make sense.

So it depends on how we want to use the `ColorPicker`, I guess. Let me know what's your preference.

components/input/ColorPicker.js
econdepe: Sounds good! I included `noop` in the new commit.

components/input/ColorPicker.js
EpokK: I see your point. For now, since we don't plan to use an input and the color picker, and because the color picker includes also a field to change the value manually, remove `useEffect` please.

econdepe: ok, done!

PR #53, opened 2019-04-25
Autocomplete

vincaslt:
### TODO

Supports multi-select:

Screenshot 2019-04-30 at 15 04 17

* [x] Add test
* [ ] Add component in `react-storybook`


components/autocomplete/Autocomplete.js
```diff
@@ -0,0 +1,104 @@
+import React, { useEffect, useRef, useState } from 'react';
+import PropTypes from 'prop-types';
+import Awesomplete from 'awesomplete';
+import SelectedItems from './SelectedItems';
+import './Autocomplete.scss';
+
+const createInputEventListener = (inputRef) => (event, callback, deps) => {
```
EpokK: This function can be move to component helper file.

components/autocomplete/Autocomplete.js
```diff
@@ -0,0 +1,104 @@
...
+            inputRef.current.removeEventListener(event, callback);
+        };
+    }, deps);
+};
+
+// TODO: move state to parent (hook?)
+const Autocomplete = ({
+    value,
+    multiple,
+    onSelect = () => {},
```
EpokK: Use `noop` coming from `proton-shared/lib/helpers/function`.

components/autocomplete/Autocomplete.js
```diff
@@ -0,0 +1,104 @@
...
+    ...rest
+}) => {
+    const inputRef = useRef(null);
+    const containerRef = useRef(null);
+    const awesompleteEventListener = createInputEventListener(inputRef);
+    const [inputValue, setInputValue] = useState(multiple ? '' : value);
+    const [selected, setSelected] = useState(value ? [].concat(value) : []);
+
+    useEffect(() => {
+        new Awesomplete(inputRef.current, {
```
EpokK: Here you need to save `Awesomplete` instance when we unmount the component and call `destroy()`.

``` js
const awesomplete = new Awesomplete();

return () => {
   awesomplete.destroy();
};
```

components/autocomplete/Autocomplete.js
```diff
@@ -0,0 +1,104 @@
...
+            ...rest,
+            container: () => containerRef.current
+        });
+    }, []);
+
+    const handleInputValueChange = (e) => {
```
EpokK: You always have `e`, so you can destructure it to get `target`.

components/autocomplete/Autocomplete.js
```diff
@@ -0,0 +1,104 @@
...
+    awesompleteEventListener('awesomplete-close', ({ reason }) => onClose(reason));
+    awesompleteEventListener('awesomplete-highlight', onHighlight);
+    awesompleteEventListener('awesomplete-open', onOpen);
+
+    return (
+        <div className="autocomplete awesomplete">
+            <div className="autocomplete-container" ref={containerRef}>
+                <div className="flex pm-field">
+                    {multiple && <SelectedItems selected={selected} onRemove={handleUnselect} />}
+                    <input
```
EpokK: You can use `Input` component maybe?

components/autocomplete/Autocomplete.js
```diff
@@ -0,0 +1,104 @@
...
+                        value={inputValue}
+                        className="w100"
+                        autoComplete="off"
+                        spellCheck={false}
```
EpokK: `spellCheck="off"` doesn't work?
```suggestion
                        spellCheck="off"
```

components/autocomplete/Autocomplete.scss
```diff
@@ -0,0 +1,22 @@
+.autocomplete input {
```
EpokK: As a good practice, try to not use nesting, define a CSS class on the input. If the component evolves and includes and other `input` this selector will break it.

components/autocomplete/Autocomplete.scss
```diff
@@ -0,0 +1,22 @@
...
+    display: flex;
+    flex: 1 0 15%;
+}
+
+.autocomplete ul {
```
EpokK: Same comment, don't use nested selector please.

components/autocomplete/Autocomplete.js
```diff
@@ -0,0 +1,104 @@
...
+
+    // TODO: cleanup this part
+    const onSelectItem = (item) => {
+        const newSelected = multiple ? [...selected, item] : [item];
+        setInputValue(multiple ? '' : item.label);
+        setSelected(newSelected);
+        onSelect(item, newSelected);
+    };
+
+    awesompleteEventListener('awesomplete-selectcomplete', ({ text }) => onSelectItem(text), [selected]);
```
EpokK: ⚠️ `useEffect` x X

components/autocomplete/SelectedItems.js
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const SelectedItems = ({ selected, onRemove }) => {
+    const handleRemove = (removed) => onRemove(removed, selected.filter((item) => item !== removed));
```
EpokK: His parent should manage the `selected` list. `onRemove` should just forward the `index` to his parent. The idea is to not have a state in `SelectedItems`, his role is just to display the list.

components/autocomplete/Autocomplete.js
vincaslt: No, it's somewhat special in that it accepts "true" or "false", and react makes it accept boolean value. https://developer.mozilla.org/en-US/docs/Mozilla/Tech/XUL/Attribute/spellcheck

components/autocomplete/Autocomplete.scss
vincaslt: For input I can, but for <ul> I don't have any control :(

components/autocomplete/Autocomplete.js
```diff
@@ -0,0 +1,144 @@
...
+
+                    <input
+                        value={inputValue}
+                        className="w100 autocomplete-input"
+                        spellCheck={false}
+                        autoComplete="off"
+                        autoCapitalize="off"
+                        onChange={handleInputValueChange}
+                        ref={inputRef}
+                        onBlur={() => handleSubmit()}
```
dhoko: No need to add the extra function here :)

components/autocomplete/Autocomplete.js
```diff
@@ -0,0 +1,144 @@
...
+                handleSubmit();
+            }}
+        >
+            <div className="autocomplete-container" ref={containerRef}>
+                <div className="flex pm-field">
+                    {selectedItems.map((item, i) => (
+                        <SelectedItem key={i} item={item} onRemove={() => onRemove(i)} />
+                    ))}
+
+                    <input
```
dhoko: Why not the default input ?

components/autocomplete/SelectedItem.js
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Icon from '../icon/Icon';
+
+const SelectedItem = ({ item, onRemove }) => {
+    const statusClasses = item.invalid ? 'autocomplete-selectedItem-invalid' : '';
+    return (
```
dhoko: What is this component about ?

components/autocomplete/useAutocomplete.js
```diff
@@ -0,0 +1,27 @@
...
+        multiple || initialSelectedItems.length === 0 ? '' : initialSelectedItems[0].label
+    );
+
+    const submit = (item) => {
+        setInputValue(multiple ? '' : item.label);
+        setSelectedItems((selected) => (multiple ? [...selected, item] : [item]));
+    };
+
+    const remove = (i) => {
+        setSelectedItems((selected) => [...selected.slice(0, i), ...selected.slice(i + 1)]);
```
dhoko: use a filter to remove it, you won't need to create copies then spread the array :)

components/autocomplete/Autocomplete.js
```diff
@@ -0,0 +1,144 @@
+import React, { useEffect, useRef, useState } from 'react';
+import PropTypes from 'prop-types';
+import Awesomplete from 'awesomplete';
+import SelectedItem from './SelectedItem';
+import './Autocomplete.scss';
+import { noop } from 'proton-shared/lib/helpers/function';
```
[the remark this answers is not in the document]
vincaslt: This is the sort of thing that should be handled by eslint. We have the plugin installed in package.json, will try looking into why it's not ordering for me automatically. Maybe the configuration is not setup?

components/autocomplete/Autocomplete.js
```diff
@@ -0,0 +1,144 @@
...
+    useEffect(() => {
+        if (awesomplete) {
+            awesomplete.list = list;
+            awesomplete.evaluate();
+        }
+    }, [awesomplete, list]);
+
+    return (
+        <form
+            className="autocomplete awesomplete"
```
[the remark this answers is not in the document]
vincaslt: I specified: `container: () => containerRef.current,` when initializing awesomplete, so it uses my container. It's not going to have 2 elements with this class.

However, I nested my div inside form instead of using form as a container, because otherwise, I must overwrite styles, that awesomplete puts on elements it injected by `awesomplete > ul`. When it's nested like I did, this selector finds nothing, and I don't need to overwrite the styles. Is there a more sensible way to do that, you think?

components/autocomplete/Autocomplete.js
vincaslt: Yeah, gonna write one handleSubmit and reuse them between form and blur.

components/autocomplete/SelectedItem.js
vincaslt: Show which values are picked from the list. I added a screenshot in the description.

components/autocomplete/Autocomplete.js
vincaslt: Input needs to support bubbles that show which values are selected. I solved this by applying input styles to a container, and the actual input is inside the container, but has no styles.

components/autocomplete/SelectedItem.js
```diff
@@ -0,0 +1,26 @@
...
+        <div className={`mr0-5 pr0-5 pl0-5 flex autocomplete-selectedItem ${statusClasses}`}>
```
EpokK: Can be useful to have an handler on double click here.

PR #55, opened 2019-04-29
Load I18n

dhoko:
- [x] Load the locale when we load the app
- [x] Auto refresh the locale
- [x] nonAuthProvider + authProvider (one doesn't need userSettings)
- [x] `useLocale` to get the locale

Need -> https://github.com/ProtonMail/proton-shared/pull/14

### DIFF
--- a/containers/Locale/Provider.js
+++ b/containers/Locale/Provider.js
@@ -1,0 +1,21 @@
+import React, { createContext, useContext, useReducer } from 'react';
+import { useUserSettings } from 'react-components';
+import { DEFAULT_TRANSLATION } from 'proton-shared/lib/constants';
+
+export const LocaleContext = createContext();
+
+export const LocaleProvider = ({ children }) => {
+    const [userSettings = { Locale: DEFAULT_TRANSLATION }] = useUserSettings();
+    const initialState = {
+        locale: userSettings.Locale
+    };
+    const reducer = (state, action) => {
+        if (action.type === 'setLocale') {
+            return { ...state, locale: action.locale };
+        }
+        return state;
+    };
+
+    return <LocaleContext.Provider value={useReducer(reducer, initialState)}>{children}</LocaleContext.Provider>;
+};
+export const useLocale = () => useContext(LocaleContext);
--- a/containers/Locale/Provider.js
+++ b/containers/Locale/Provider.js
@@ -1,5 +1,5 @@
 import { c } from 'ttag';
-import React, { createContext, useContext, useState } from 'react';
+import React, { createContext, useContext, useState, useEffect } from 'react';
 import PropTypes from 'prop-types';
 import { useUserSettings, useConfig, useNotifications } from 'react-components';
 import { DEFAULT_TRANSLATION } from 'proton-shared/lib/constants';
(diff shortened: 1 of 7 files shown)


containers/Locale/Provider.js
```diff
@@ -0,0 +1,38 @@
...
+
+    // Load the translations when we load the app
+    useEffect(() => {
+        loadLocale(config, Locale).then(() => {
+            setState(Locale); // force refresh children
+        });
+    }, []);
+
+    // Only when the locale changes via event
+    if (state !== Locale) {
```
mmso: Don't do side effects in the render function, put it in a `useEffect`.

containers/Locale/Provider.js
```diff
@@ -0,0 +1,38 @@
...
+        loadLocale(config, Locale).then(() => {
+            setState(Locale); // force refresh children
+            createNotification({ text: c('Success').t`Locale updated` });
+        });
+    }
+
+    return <LocaleContext.Provider value={state}>{children}</LocaleContext.Provider>;
```
mmso: Does it really need to provide the current locale?

It can be calculated from settings model right?

containers/cache/ModelsProvider.js
```diff
@@ -14,19 +14,19 @@ const ModelsProvider = ({ children, loginData = {} }) => {
...
 
         const setup = async () => {
-            const [user, eventID, userSettingsModel, mailSettingsModel] = await Promise.all([
+            const userSettingsModel = await UserSettingsModel.get(api);
+            const [user, eventID, mailSettingsModel] = await Promise.all([
                 loginData.user || UserModel.get(api),
                 loginData.eventID || getEventID(api),
-                UserSettingsModel.get(api),
                 MailSettingsModel.get(api)
+                // loadLocale(config, userSettingsModel.Locale)
```
mmso: This shouldn't be here anymore

For the initial loading spinner we need to come up with another solution

I think we need two things: 
* one initial loading state that just does the user request to check if the user is logged in or not
* and if logged in, a second loading state that resolves whatever the application needs

containers/Locale/Provider.js
dhoko: What do you mean ?

containers/cache/ModelsProvider.js
dhoko: > one initial loading state that just does the user request to check if the user is logged in or not

That's why we use the provider. If logged-in or not doesn't really matter.

containers/Locale/Provider.js
mmso: Why does it need to provide the locale when it can be retrieved from the settings model?

dhoko: Hum imho it should not be inside Models provider.

containers/cache/ModelsProvider.js
mmso: What happens when the user logins with a new locale and the request to fetch the locale takes a long time?

If it displays the UI before the request to fetch the locale has finished I think it's better to show a loading spinner. However this is something we need to add for the dynamic models anyway, and it can be done in a later step.

containers/Locale/Provider.js
mmso: Why? Every other settings component is reading values from the model

dhoko: because it's the locale

containers/cache/ModelsProvider.js
```diff
@@ -14,18 +14,17 @@ const ModelsProvider = ({ children, loginData = {} }) => {
...
-
     const [{ loading, error }, setState] = useState({ loading: true });
 
     useEffect(() => {
         const abortController = new AbortController();
         const apiWithAbort = (config) => api({ ...config, signal: abortController.signal });
 
         const setup = async () => {
-            const [user, eventID, userSettingsModel, mailSettingsModel] = await Promise.all([
+            const userSettingsModel = await UserSettingsModel.get(api);
```
mmso: can revert this now?

dhoko: yup :+1:

PR #62, opened 2019-05-07
fix all modals

### LINKED ISSUE #58: Fix Modal to use new pattern
The `state` is attach to the modal and persist even if the modal is not display.

### To do
* [ ] Remove `show` on all Modals
* [ ] Use the new pattern
 
Here is the new pattern:

``` jsx
{isOpen ? <Modal /> : null}
```

econdepe:
Closes #58

### DIFF
--- a/components/modal/Confirm.js
+++ b/components/modal/Confirm.js
@@ -7,9 +7,9 @@ import Modal from './Modal';
 import Footer from './Footer';
 import Content from './Content';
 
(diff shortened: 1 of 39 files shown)


econdepe:
> You missed `PreviewPDFModal` in `InvoiceActions`. The rest is OK for me.

Arggh, I forgot to double-check and search for `'show='`...

PR #63, opened 2019-05-08
Components for auto-reply feature

vincaslt:
Auto-reply functionality for: https://github.com/ProtonMail/protonmail-settings/issues/10

**Related PRs:**

- `protonmail-settings`: https://github.com/ProtonMail/protonmail-settings/pull/29
- `design-system`: https://github.com/ProtonMail/design-system/pull/114 ✅
- `proton-shared`: https://github.com/ProtonMail/proton-shared/pull/16 ✅

Screenshot 2019-05-08 at 16 32 12

### DIFF
--- a/components/input/RichTextEditor.js
+++ b/components/input/RichTextEditor.js
@@ -1,5 +1,10 @@
 import React from 'react';
 import ReactQuill from 'react-quill';
+import Quill from 'quill';
+
+const Block = Quill.import('blots/block');
+Block.tagName = 'div';
+Quill.register(Block);
 
 const RichTextEditor = ({ ...rest }) => <ReactQuill {...rest} />;
 
--- a/components/select/TimeSelect.js
+++ b/components/select/TimeSelect.js
@@ -1,0 +1,35 @@
+import React from 'react';
+import { range } from 'proton-shared/lib/helpers/array';
+import { noop } from 'proton-shared/lib/helpers/function';
+import PropTypes from 'prop-types';
+import Select from './Select';
+
+const HOUR = 60 * 60 * 1000;
+
+const timeSelectOptions = range(0, 24).reduce((options, hour) => {
+    const formattedHour = hour < 10 ? `0${hour}` : hour;
+    const hours = hour * HOUR;
+    return [
+        ...options,
+        { text: `${formattedHour}:00`, value: hours },
+        { text: `${formattedHour}:30`, value: hours + HOUR / 2 }
+    ];
+}, []);
+
+const TimeSelect = ({ onChange, value, ...rest }) => {
+    const handleChange = (e) => onChange(parseInt(e.target.value, 10));
+
+    return <Select {...rest} options={timeSelectOptions} value={value} onChange={handleChange} />;
+};
+
+TimeSelect.propTypes = {
+    value: PropTypes.number,
+    onChange: PropTypes.func
+};
+
+TimeSelect.defaultValues = {
+    value: 0,
+    onChange: noop
+};
+
+export default TimeSelect;
--- a/components/select/TimeSelect.js
+++ b/components/select/TimeSelect.js
@@ -17,7 +17,7 @@ const timeSelectOptions = range(0, 24).reduce((options, hour) => {
 }, []);
 
 const TimeSelect = ({ onChange, value, ...rest }) => {
-    const handleChange = (e) => onChange(parseInt(e.target.value, 10));
+    const handleChange = ({ target }) => onChange(+target.value);
 
     return <Select {...rest} options={timeSelectOptions} value={value} onChange={handleChange} />;
 };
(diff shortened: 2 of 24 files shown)


components/input/RichTextEditor.js
```diff
@@ -1,5 +1,10 @@
 import React from 'react';
 import ReactQuill from 'react-quill';
+import Quill from 'quill';
+
+const Block = Quill.import('blots/block');
```
EpokK: Yeah, I think for now that's OK and I'm not seeing a case where `p` is useful for us.

containers/autoReply/AutoReplyTemplate.js
```diff
@@ -0,0 +1,85 @@
...
+                    <InfoLine label={c('Label').t`Duration`}>{durationLabel}</InfoLine>
+                    <InfoLine label={c('Label').t`Start`}>{formatTime(autoresponder.StartTime * 1000)}</InfoLine>
+                    <InfoLine label={c('Label').t`End`}>{formatTime(autoresponder.EndTime * 1000)}</InfoLine>
+                    <InfoLine label={c('Label').t`Timezone`}>{timezone}</InfoLine>
+                    <InfoLine plain label={c('Label').t`Message`}>
+                        <div dangerouslySetInnerHTML={{ __html: autoresponder.Message }} />
+                    </InfoLine>
+                </tbody>
+            </table>
+            <div className="flex w100">
```
EpokK: Why do you need `flex w100` here?

components/select/TimeSelect.js
```diff
@@ -0,0 +1,35 @@
...
+    const hours = hour * HOUR;
+    return [
+        ...options,
+        { text: `${formattedHour}:00`, value: hours },
+        { text: `${formattedHour}:30`, value: hours + HOUR / 2 }
+    ];
+}, []);
+
+const TimeSelect = ({ onChange, value, ...rest }) => {
+    const handleChange = (e) => onChange(parseInt(e.target.value, 10));
```
EpokK:
```suggestion
    const handleChange = ({ target }) => onChange(+target.value);
```

containers/autoReply/AutoReplyModal.js
```diff
@@ -0,0 +1,48 @@
...
+
+const AutoReplyModal = ({ onClose }) => {
+    const [{ AutoResponder }] = useMailSettings();
+    const { model, updateModel, toAutoResponder } = useAutoReplyForm(AutoResponder);
+    const { request } = useApiWithoutResult(updateAutoresponder);
+    const { call } = useEventManager();
+
+    const handleSubmit = async () => {
+        await request(toAutoResponder(model));
+        call();
```
EpokK: We should display a notification to say "Auto-reply updated". For notification you can use `useNotifications`.

``` js
const { createNotification } = useNotifications();
createNotification({ text: c('Success').t`Auto-reply updated` });
```

containers/autoReply/AutoReplyForm/useAutoReplyForm.js
```diff
@@ -0,0 +1,66 @@
...
+    DaysSelected: model.daysOfWeek,
+    Zone: model.timeZone,
+    Subject: model.subject,
+    IsEnabled: model.enabled,
+    StartTime: (model.startDate + model.startTime) / 1000,
+    EndTime: (model.endDate + model.endTime) / 1000
+});
+
+const useAutoReplyForm = (AutoResponder) => {
+    const [model, setModel] = useState(() => toModel(AutoResponder));
```
EpokK: Why the model is a `function`? Why about an `object`?

components/input/RichTextEditor.js
vincaslt: Would be nice to find a better way to set defaults for this component. Now this sets div tag as a default element for every instance of RichTextEditor. Might be useful to allow configuring individually.

containers/autoReply/AutoReplyForm/useAutoReplyForm.js
vincaslt: My model is an object. When a function is passed to useState - it only calls it once on initial render. If I wrote `useState(toModel(AutoResponder));` my toModel function would get called every time the component re-rendered.

https://reactjs.org/docs/hooks-reference.html#lazy-initial-state

containers/autoReply/AutoReplyTemplate.js
```diff
@@ -0,0 +1,83 @@
...
+            ).text;
+            return c('Duration').t`Every ${dayOfMonth} @ ${hours}`;
+        } else if (autoresponder.Repeat === AutoReplyDuration.PERMANENT) {
+            return c('Duration').t`n/a`;
+        }
+    };
+
+    return (
+        <div className="bordered-container p2 mw650p">
+            <table>
```
EpokK: Why not using a structure with `Row`, `Label` and `Field`?

vincaslt: From the design it seemed that the all of the labels must have the same width and take it from the widest one. That's basically a table. If I use Rows, labels would have different sizes or I would need to set a fixed size for them. I've hard opinion on this, just it seemed to be simpler this way.

PR #65, opened 2019-05-08
Add animation when we close a modal

EpokK:
When closing the modal, the class `pm-modalOverlay--fadeOut` is added on `pm-modalOverlay`, let this animation execute itself before removing the code in the DOM (500ms duration).

### DIFF
--- a/components/modal/useModal.js
+++ b/components/modal/useModal.js
@@ -8,13 +8,19 @@ const useModal = (initialState = false) => {
 
     const close = () => {
         const modal = document.querySelector('.pm-modalOverlay');
-        modal && modal.classList.add('pm-modalOverlay--fadeOut');
-        id = setTimeout(() => setModalState(false), 500);
+
+        if (modal) {
+            modal.classList.add('pm-modalOverlay--fadeOut');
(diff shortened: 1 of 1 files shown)


components/modal/useModal.js
```diff
@@ -1,10 +1,28 @@
...
 const useModal = (initialState = false) => {
     const [isOpen, setModalState] = useState(initialState);
-    const open = () => setModalState(true);
-    const close = () => setModalState(false);
     const toggle = () => setModalState(!isOpen);
+    const open = () => setModalState(true);
+    let id;
+
+    const close = () => {
+        const modal = document.querySelector('.pm-modalOverlay');
```
vincaslt: What about some very rare cases when there are two popups visible? The one I've seen was when my session ends while a popup is opened, and I get another popup asking me to enter my password again. This selector seems like it might fade out all of the visible popups.

vincaslt: Screenshot 2019-05-09 at 09 51 46

EpokK:
We will revamp modal to add this feature.

Issue #59, opened 2019-05-02
Use dropdown design


EpokK:
https://design-system-beta.netlify.com/drop-down/

PR #71, opened 2019-05-10
Fix dropdown styling

mmso:
Changes:

* `DropdownActions` now only supports a list of buttons, since the primary button can only be a button
* Adds support for animation
* `DropdownMenu` now only supports children, instead of receiving a list as array.
* Removes `DropdownItem` because I want to avoid generic components if they end up taking a lot of props. Replaced by `DropdownButton`, `DropdownLink`, etc., or whatever you want.

How to use:

```jsx
            <Dropdown
                align="[right|left]"
                caret
                className={'pm-button'}
                title={'Open dropdown'}
                content={'Press me'}
            >
                <DropdownMenu>
                    <DropdownButton className="alignleft">
                       First button
                    </DropdownButton>
                    <DropdownButton className="alignleft" onClick={handleClick}>
                       Second button
                    </DropdownButton>
                    <DropdownLink className="alignleft" href="http://www.foo.bar">
                       A link
                    </DropdownLink>
                </DropdownMenu>
                <PrimaryButton>Another button below the menu</PrimaryButton>
            </Dropdown>
```

Closes: #59


containers/domains/DomainAddresses.js
```diff
@@ -7,7 +7,7 @@ import { Dropdown, DropdownMenu, Icon } from 'react-components';
 const DomainAddresses = ({ domain }) => {
     const addresses = domain.addresses || [];
     const title = addresses.map(({ Email }) => Email).join(', ');
-    const list = addresses.map(({ Email: text }) => ({ text }));
+    const actions = addresses.map(({ Email: text }) => text);
```
dhoko: useless alias ;)

containers/keys/KeysActions.js
```diff
@@ -46,11 +46,13 @@ const ACTIONS_TO_TEXT = {
...
-        text: ACTIONS_TO_TEXT[action]().text,
-        type: 'button',
-        onClick: cb
-    }));
+    const list = actions
+        .map(({ action, cb }) => ({
+            text: ACTIONS_TO_TEXT[action]().text,
+            onClick: cb
+        }))
+        .filter(Boolean);
```
dhoko: useless filter the array is always full

containers/members/MemberAddresses.js
```diff
@@ -7,8 +7,8 @@ import { Dropdown, DropdownMenu, Icon } from 'react-components';
 const MemberAddresses = ({ member }) => {
     const addresses = member.addresses || [];
     const title = addresses.map(({ Email }) => Email).join(', ');
-    const list = addresses.map(({ Email: text }) => ({ text }));
-    const n = list.length;
+    const actions = addresses.map(({ Email: text }) => text);
```
dhoko: idem here for the alias ;)

components/dropdown/Dropdown.js
```diff
@@ -44,16 +59,27 @@ const Dropdown = ({ isOpen, children, className, content, autoClose, autoCloseOu
...
-            <Button className={className} onClick={handleClick} aria-expanded={open}>
-                {content}
-            </Button>
-            {open ? (
-                <div className="dropDown-content" onClick={handleClickContent}>
-                    {children}
-                </div>
-            ) : null}
+        <div className={`${dropdownClassName} ${className}`} ref={wrapperRef}>
+            <button
```
EpokK: I suppose you have a reason to not use the `Button` component here. Then did you check how make the dropdown looks primary if needed?

components/dropdown/Dropdown.js
```diff
@@ -44,16 +59,27 @@ const Dropdown = ({ isOpen, children, className, content, autoClose, autoCloseOu
...
+                className="increase-surface-click"
+                aria-expanded={open}
+                onClick={handleClick}
+                type="button"
+            >
+                <span className="mauto">
+                    {content} {caretContent}
+                </span>
+            </button>
+            <div className={contentClassName} onClick={handleClickContent} hidden={!open}>
```
EpokK: Why keep the content of the `Dropdown` in the page? I suppose for animation?

mmso: Yes, and we can attach an event handler for `onAnimationEnd` and remove it if we see that it becomes a performance issue to always render all of them. But I think it should be ok so I didn't add it for now

PR #75, opened 2019-05-16
[UI] Page filters + upgrade modals + move from protonmail-settings

dhoko:


components/Filters/editor/Actions.js
```diff
@@ -0,0 +1,146 @@
...
+    };
+
+    const handleOnChangeLabel = (labels) => {
+        onChange(ACTIONS.labels(labels));
+    };
+
+    return (
+        <Row>
+            <Label htmlFor="actions">{c('New Label form').t`Actions`}</Label>
+            <div className="w100">
```
EpokK:
```suggestion
            <Field className="w100">
```

components/Filters/editor/Actions.js
```diff
@@ -0,0 +1,146 @@
...
+                        onChange={handleChange('markAs')}
+                        className="mlauto"
+                        defaultValue={getDefaultValue('markAs')}
+                    />
+                </Row>
+
+                {errors.isValid === false ? (
+                    <ErrorZone id="ActionsError">{c('Error').t`A filter must have an action`}</ErrorZone>
+                ) : null}
+            </div>
```
EpokK:
```suggestion
            </Field>
```

containers/Labels/modals/Edit.js
```diff
@@ -0,0 +1,108 @@
...
+
+    const [model, setModel] = useState(
+        label || {
+            Name: '',
+            Color: LABEL_COLORS[randomIntFromInterval(0, LABEL_COLORS.length - 1)],
+            Exclusive: +(props.type === 'folder')
+        }
+    );
+
+    useEffect(() => {
```
EpokK: Why do you need to set the `model` when the model change? It seems useless right?

components/Filters/ActionsFilterToolbar.js
```diff
@@ -0,0 +1,24 @@
...
+    const { createModal } = useModals();
+
+    const handleClickAdd = (type) => () => {
+        createModal(<AddFilterModal type={type} />);
+    };
+
+    return (
+        <>
+            <PrimaryButton onClick={handleClickAdd()}>{c('Action').t`Add Filter`}</PrimaryButton>
+            <Button onClick={handleClickAdd('complex')} className="ml1">
```
EpokK: As a good practice use `mr1` on the first button than the second button.

containers/Filters/AddFilterModal.js
```diff
@@ -0,0 +1,196 @@
...
+import ConditionsEditor from '../../components/Filters/editor/Conditions';
+import ActionsEditor from '../../components/Filters/editor/Actions';
+import OperatorEditor from '../../components/Filters/editor/Operator';
+import SieveEditor from '../../components/Filters/editor/Sieve';
+import PreviewFilter from '../../components/Filters/editor/Preview';
+import NameEditor from '../../components/Filters/editor/Name';
+
+import './AddFilterModal.css';
+
+function AddFilterModal({ filter, type, mode, onEdit, ...props }) {
```
EpokK: For me it makes more sense to have `mode` for `simple` or `complex`. And `method` for `create` or `update`.

components/Filters/FilterConditionValues.js
```diff
@@ -0,0 +1,62 @@
...
+function FilterConditionValues({ condition, options, onDelete, onChangeCondition, onAdd, onEdit, error }) {
+    const hasError = (key) => (error.errors || []).includes(key);
+    return (
+        <>
+            <Select
+                options={options}
+                name="filterConditions"
+                className="mb1"
+                onChange={onChangeCondition}
+                defaultValue={condition.Comparator.value}
```
EpokK: What is the `defaultValue`? Even if you need to select a default value for the select you should manage that with his `value` when you initialize his state.

components/Filters/FilterItemRow.js
```diff
@@ -0,0 +1,59 @@
...
+import { isComplex } from 'proton-shared/lib/filters/factory';
+import { toggleEnable } from 'proton-shared/lib/api/filters';
+import { Icon, Toggle, useEventManager, useApiWithoutResult, useNotifications } from 'react-components';
+
+import RemoveFilter from './RemoveFilter';
+import EditFilterButton from './EditFilterButton';
+
+function FilterItemRow({ filter }) {
+    const { ID, Name, Status } = filter;
+    const [toggled, setToggle] = useState(Status === 1);
```
EpokK: Use a constant if for the status please.

components/Filters/FilterItemRow.js
```diff
@@ -0,0 +1,59 @@
...
+                <Icon name="text-justify" />
+            </td>
+            <td>{Name}</td>
+            <td>
+                <div className="w10">
+                    <Toggle
+                        id={`item-${ID}`}
+                        loading={loading}
+                        checked={toggled}
+                        onChange={debounce(handleChangeStatus, 300)}
```
EpokK: Why using a `debounce` here?

components/Filters/FilterItemRow.js
EpokK: Use `useToggle`.

components/Filters/FilterItemRow.js
```diff
@@ -0,0 +1,59 @@
...
+    const { ID, Name, Status } = filter;
+    const [toggled, setToggle] = useState(Status === 1);
+
+    const { call } = useEventManager();
+    const { createNotification } = useNotifications();
+    const { request, loading } = useApiWithoutResult(toggleEnable);
+
+    const handleChangeStatus = async () => {
+        await request(ID, !toggled);
+        setToggle(!toggled);
```
EpokK: Use `toggle` function coming from `useToggle`.

components/Filters/FilterItemRow.js
```diff
@@ -0,0 +1,59 @@
...
+function FilterItemRow({ filter }) {
+    const { ID, Name, Status } = filter;
+    const [toggled, setToggle] = useState(Status === 1);
+
+    const { call } = useEventManager();
+    const { createNotification } = useNotifications();
+    const { request, loading } = useApiWithoutResult(toggleEnable);
+
+    const handleChangeStatus = async () => {
+        await request(ID, !toggled);
```
EpokK: Use the value coming from `event.target.checked`.

components/Filters/RadioContainsAttachements.js
```diff
@@ -0,0 +1,55 @@
+import React, { useState, useEffect } from 'react';
+import PropTypes from 'prop-types';
+import { c } from 'ttag';
+import { Radio, Row } from 'react-components';
+import { noop } from 'proton-shared/lib/helpers/function';
+
+function RadioContainsAttachements({ comparator, onChange }) {
+    const [value, setValue] = useState(comparator);
+    useEffect(() => {
```
EpokK: Why setting the `value` when the `value` change?

components/Filters/RadioContainsAttachements.js
```diff
@@ -0,0 +1,55 @@
...
+        setValue(value);
+    }, [value]);
+    const handleChange = (e) => {
```
EpokK:
```suggestion
    const handleChange = ({ target }) => {
```

components/Filters/RadioContainsAttachements.js
```diff
@@ -0,0 +1,55 @@
+import React, { useState, useEffect } from 'react';
+import PropTypes from 'prop-types';
+import { c } from 'ttag';
+import { Radio, Row } from 'react-components';
+import { noop } from 'proton-shared/lib/helpers/function';
+
+function RadioContainsAttachements({ comparator, onChange }) {
+    const [value, setValue] = useState(comparator);
```
EpokK: Do you really need a `state` here since you have one also [outside](https://github.com/ProtonMail/react-components/pull/75/files#diff-881f831d03e0a620e6d31f1f4980efb2R113)?

components/Filters/FilterItemRow.js
```diff
@@ -0,0 +1,59 @@
...
+                        loading={loading}
+                        checked={toggled}
+                        onChange={debounce(handleChangeStatus, 300)}
+                    />
+                </div>
+            </td>
+            <td>
+                {!isComplex(filter) ? <EditFilterButton filter={filter} textContent={c('Action').t`Edit`} /> : null}
+                <EditFilterButton filter={filter} type="complex" textContent={c('Action').t`Edit Sieve`} />
+                <RemoveFilter filter={filter} />
```
EpokK: You are not using `onRemoveFilter`?

components/Filters/RemoveFilter.js
```diff
@@ -0,0 +1,55 @@
...
+
+function RemoveFilter({ filter, className, onRemoveFilter }) {
+    const { call } = useEventManager();
+    const { createModal } = useModals();
+    const { createNotification } = useNotifications();
+    const { request, loading } = useApiWithoutResult(deleteFilter);
+
+    const handleConfirmConfirmModal = async () => {
+        await request(filter.ID);
+        call();
```
EpokK: If you are not using `onRemoveFilter` when you use this component you should wait on the event manager to refresh the filter collection before displaying the success notification.
```suggestion
        await call();
```

EpokK: Also if the API call for the event manager fails, the user will still see the success notification and the collection will not be updated on the UI.

components/Filters/SortableList.js
```diff
@@ -0,0 +1,42 @@
+import React from 'react';
+import { c } from 'ttag';
+import { SortableContainer } from 'react-sortable-hoc';
+import { Icon } from 'react-components';
+
+import FilterItemRow from './FilterItemRow';
+
+export default SortableContainer(({ items, onClickEdit, onRemoveFilter, onChangeStatus }) => {
+    return (
+        <table className="pm-simple-table noborder border-collapse mt1">
```
EpokK: You can probably use `Table` component here.

components/Filters/editor/Preview.js
```diff
@@ -0,0 +1,146 @@
...
+                    value: labels
+                });
+        }
+
+        return acc;
+    }, []);
+};
+
+function PreviewFilter({ filter }) {
+    const [labelsModel] = useFormattedLabels();
```
EpokK:
```suggestion
    const [labelsModel = []] = useFormattedLabels();
```

containers/Filters/Container.js
```diff
@@ -0,0 +1,66 @@
...
+
+    useEffect(() => {
+        setFilters(filters || []);
+    }, [filters]);
+
+    const getScrollContainer = () => document.querySelector('.main-area');
+    const onSortEnd = async ({ oldIndex, newIndex }) => {
+        const newList = arrayMove(list, oldIndex, newIndex);
+        setFilters(newList);
+        await orderRequest.request(newList.map(({ ID }) => ID));
```
EpokK: If the request fails, you need to reset the filter list from his previous state.

components/Filters/spamlist/SearchEmailIntoList.js
```diff
@@ -0,0 +1,38 @@
...
+    const handleSeachChange = async (Keyword) => {
+        onBeforeRequest(Keyword);
+        const { IncomingDefaults = [] } = await request({ Keyword, PageSize: 200 });
+        onAfterRequest(Keyword, IncomingDefaults);
+    };
+
+    return (
+        <SearchInput
+            className={'w100 '.concat(className)}
+            onChange={debounce(handleSeachChange, 300)}
```
EpokK: You can use the `delay` property instead of the `debounce` here, example:
``` jsx
<SearchInput delay={500} onChange={handleChange} value={keywords} />
```

components/Filters/spamlist/SearchEmailIntoList.js
```diff
@@ -0,0 +1,38 @@
...
+import PropTypes from 'prop-types';
+import { c } from 'ttag';
+import { SearchInput, useApiWithoutResult } from 'react-components';
+import { getIncomingDefaults } from 'proton-shared/lib/api/incomingDefaults';
+import { noop, debounce } from 'proton-shared/lib/helpers/function';
+
+function SearchEmailIntoList({ className, onBeforeRequest, onAfterRequest }) {
+    const { request } = useApiWithoutResult(getIncomingDefaults);
+
+    const handleSeachChange = async (Keyword) => {
```
EpokK: `handleSeachChange` => `handleSearchChange`

components/Filters/spamlist/SearchEmailIntoList.js
```diff
@@ -0,0 +1,38 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { c } from 'ttag';
+import { SearchInput, useApiWithoutResult } from 'react-components';
+import { getIncomingDefaults } from 'proton-shared/lib/api/incomingDefaults';
+import { noop, debounce } from 'proton-shared/lib/helpers/function';
+
+function SearchEmailIntoList({ className, onBeforeRequest, onAfterRequest }) {
```
EpokK: `onBeforeRequest` and `onAfterRequest` should be handle by the parent component to handle state change there.

```suggestion
function SearchEmailIntoList({ className, value, onChange }) {
```

components/Filters/spamlist/SearchEmailIntoList.js
```diff
@@ -0,0 +1,38 @@
...
+    const { request } = useApiWithoutResult(getIncomingDefaults);
+
+    const handleSeachChange = async (Keyword) => {
+        onBeforeRequest(Keyword);
+        const { IncomingDefaults = [] } = await request({ Keyword, PageSize: 200 });
```
EpokK: ⚠️ if the request fails, you need to hide loaders.

components/Filters/editor/Preview.js
dhoko: The hook will always return factory from an Array ;)

containers/Filters/AddFilterModal.css
```diff
@@ -0,0 +1,3 @@
+.AddFilterModal-isPreview .AddFilterModal-editor {
```
EpokK: Why not using a ternary condition inside the React component?

containers/Filters/SpamContainer.js
```diff
@@ -0,0 +1,87 @@
...
+    const handleAction = (action) => (type, data) => {
+        action === 'create' && create(type, data);
+        action === 'remove' && remove(data);
+        action === 'move' && move(type, data);
+    };
+
+    return (
+        <>
+            <SubTitle>{c('FilterSettings').t`Spam Filters`}</SubTitle>
+            <p className="block-info-standard mt1 mb1">
```
EpokK: Use `Alert` with his prop `learnMore`.

components/Filters/FilterItemRow.js
dhoko: To prevent too many requests
cf (with, then without the debounce)

output

components/Labels/LabelColors.js
```diff
@@ -0,0 +1,49 @@
...
+                    >
+                        <Input
+                            type="radio"
+                            onChange={onChange(color)}
+                            value={color}
+                            name="paletteColor"
+                            aria-labelledby={`Color ${color}`}
+                            className="LabelColors-input-color"
+                        />
+                        <div className="LabelColors-item-mask flex">
```
EpokK: Do you need `flex` here?

components/Filters/ActionsFilterToolbar.js
dhoko: >  The second element should care about the previous one.

yep that's why there is the className on it ;) cf the adjacent selector

containers/Filters/AddFilterModal.css
dhoko: Because it's easier to do it once ;) on the container

components/Labels/ToggleNotify.js
```diff
@@ -0,0 +1,35 @@
+import React, { useState } from 'react';
+import PropTypes from 'prop-types';
+import { c } from 'ttag';
+import { debounce } from 'proton-shared/lib/helpers/function';
+import { Toggle, useApiWithoutResult, useEventManager, useNotifications } from 'react-components';
+import { updateLabel } from 'proton-shared/lib/api/labels';
+
+const ToggleNotify = ({ label }) => {
+    const [toggled, setToggle] = useState(label.Notify === 1);
```
EpokK: Use `useToggle`.

components/Labels/ToggleNotify.js
```diff
@@ -0,0 +1,35 @@
...
+    const { call } = useEventManager();
+    const { createNotification } = useNotifications();
+    const { request, loading } = useApiWithoutResult(updateLabel);
+
+    const handleChange = async () => {
+        const newLabel = {
+            ...label,
+            Notify: +!label.Notify
```
EpokK: Use the value coming from `event.target.checked`, it's safer if `label.Notify` change from somewhere.

components/Labels/ToggleNotify.js
```diff
@@ -0,0 +1,35 @@
...
+        };
+        await request(label.ID, newLabel);
+        setToggle(newLabel.Notify === 1);
+        call();
+        createNotification({
+            text: c('label/folder notification').t`${label.Name} updated`
+        });
+    };
+    return (
+        <Toggle id={`item-${label.ID}`} checked={toggled} onChange={debounce(handleChange, 300)} loading={loading} />
```
EpokK: Why are you using a `debounce` here?

components/Filters/FilterItemRow.js
EpokK: But the `loading` prop should block this behavior when the API request is running.

components/Labels/ToggleNotify.js
dhoko: It's null and as it's a toggle !value works (if label changes, the toggle will too :) )

components/Filters/spamlist/MoveEmailFilteredList.js
```diff
@@ -0,0 +1,53 @@
...
+
+    const I18N = {
+        blacklist: c('Title').t`blacklist`,
+        whitelist: c('Title').t`whitelist`
+    };
+
+    const handleClick = async () => {
+        const { Email, ID } = email;
+        const Location = dest === 'whitelist' ? WHITELIST_TYPE : BLACKLIST_TYPE;
+        const { IncomingDefault: data } = await request(ID, { Location });
```
mmso: I think the logic of this function would be better handled in `SpamFiltersContainer` because it would allow to set the temporary list before starting the request / and reset it if it fails etc.

and it could return a promise and to disable this button while the action is running

components/Filters/editor/Preview.js
```diff
@@ -0,0 +1,146 @@
+import React from 'react';
+import { c } from 'ttag';
+import PropTypes from 'prop-types';
+import { Alert, Row, Label, useFormattedLabels } from 'react-components';
+
+import PreviewActionValue from './PreviewActionValue';
+import './Preview.css';
+
+const fomartActions = (actions, labelsModel) => {
```
mmso:
```suggestion
const formatActions = (actions, labelsModel) => {
```

components/Labels/ToggleNotify.js
EpokK: The `Toggle` is a checkbox, so `checked` should be defined if you use `checked` property on the `Toggle`. It's always better to use the value coming from the event than the current state. Think the component as a controlled input.

components/Filters/spamlist/MoveEmailFilteredList.js
dhoko: pagination comes from the API so we can't use do a temp list as 95% of the time for the user it's the same as seeing a delete action.

containers/Filters/AddFilterModal.js
```diff
@@ -0,0 +1,196 @@
...
+                !isPreview ? c('Add Filter Modal').t`Custom Filter` : c('Add Filter Modal').t`Custom Filter (Preview)`
+            }
+            onClose={props.onClose}
+            footer={getFooter(reqCreate.loading || reqUpdate.loading)}
+            {...props}
+        >
+            {type === 'complex' ? (
+                <div className="AddFilterModal-editor">
+                    <NameEditor error={errors.name} filter={filterModel} onChange={handleChangeName} />
+                    <SieveEditor
```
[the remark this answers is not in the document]
dhoko: ping ?

containers/Filters/AddFilterModal.js
dhoko: method too technical imho. Mode is a behavior.

package.json
```diff
@@ -45,8 +46,11 @@
     "push.js": "^1.0.9",
     "qrcodejs2": "0.0.2",
     "quill": "^1.3.6",
+    "react-codemirror2": "^6.0.0",
     "react-color": "^2.17.0",
-    "react-quill": "^1.3.3"
+    "react-quill": "^1.3.3",
+    "react-sortable-hoc": "^1.9.1",
+    "tooltip.js": "^1.3.1"
```
EpokK: We are not using `tooltip.js` anymore.

containers/Filters/useSpamList.js
```diff
@@ -0,0 +1,137 @@
...
+    if (!input) {
+        return defaultFilter;
+    }
+    return ({ Email }) => Email.toLowerCase().includes(input.toLowerCase());
+};
+
+const useSpamList = () => {
+    const [whiteList, setWhiteList] = useState([]);
+    const [blackList, setBlackList] = useState([]);
+    const [searchQuery, setSearchQuery] = useState('');
```
EpokK: Use `useSearch` hook.

components/Filters/RemoveFilter.js
```diff
@@ -0,0 +1,55 @@
...
+        await call();
+        createNotification({
+            text: c('Filter notification').t`Filter removed`
+        });
+        onRemoveFilter(filter);
+    };
+
+    const handleClick = () =>
+        createModal(
+            <ConfirmModal loading={loading} onConfirm={handleConfirmConfirmModal} title={c('Title').t`Delete Filter`}>
```
EpokK: This `loading` is useless because we close the modal when the user click on confirm. The loading state should be present on the action which open the modal.

components/Filters/editor/Conditions.js
```diff
@@ -0,0 +1,187 @@
...
+                                <Label>
+                                    {c('Label').t`Conditions ${index + 1}`}
+                                    <SmallButton className="ml1" onClick={handleRemoveCondition(index)}>
+                                        <Icon name="trash" />
+                                    </SmallButton>
+                                </Label>
+
+                                <div className="w100">
+                                    <Row>
+                                        <Select options={toOptions(TYPES)} defaultValue={condition.Type.value} />
```
EpokK: No listener?

components/Filters/editor/LabelActions.js
```diff
@@ -0,0 +1,64 @@
...
+            onChange(selection.map(({ Name }) => Name));
+        }
+    });
+
+    return (
+        <>
+            <input />
+            <Autocomplete
+                placeholder={c('Placeholder').t`Add a label ...`}
+                className="w100"
```
EpokK: The `Autocomplete` input should already have the `w100` class. If it's not the case, you can add it please?

components/Filters/editor/LabelActions.js
```diff
@@ -0,0 +1,64 @@
...
+            />
+
+            <ul className="m0 mt1 mb1 p0">
+                {selectedItems.map(({ Color, Name }, i) => {
+                    const className = 'flex '.concat(i ? 'mt0-5' : '');
+                    return (
+                        <li key={Name} className={className}>
+                            <Icon name="label" style={{ fill: Color }} className="mr1" alt={Name} />
+                            <span>{Name}</span>
+                            <SmallButton type="button" onClick={() => deselect(i)} className="mlauto">{c('Action')
```
EpokK: We already force `type="button"` on `Button` component.

EpokK: I let you do a search in your PR for other cases.

components/Filters/spamlist/RemoveEmailFilteredList.js
```diff
@@ -0,0 +1,46 @@
...
+        await request([ID]);
+        createNotification({
+            text: c('Moved to black/whitelist').t`${Email} removed from ${I18N[type]}`
+        });
+        onClick(type, email);
+    };
+
+    return (
+        <>
+            <SmallButton className={className} onClick={handleClick}>
```
EpokK: You should use a `loading` state on the button to not run the request several times.

components/Filters/ActionsFilterToolbar.js
```diff
@@ -0,0 +1,24 @@
...
+                {c('Action').t`Add Sieve Filter`}
```
EpokK: Don't hover capitalize text please. I know on Zeplin it's the case, but it's a mistake ...

components/Filters/FilterItemRow.js
EpokK: ?

dhoko: onRemoveFilter doesn't exist :) and as we use the event it was useless.

PR #80, opened 2019-05-21
Add ability to bind onChange to autocomplete hook

vincaslt:
proton-shared PR for identity function: https://github.com/ProtonMail/proton-shared/pull/24

Initial idea was that if you wanted to keep track of selected/removed events in components using autocomplete hook, you would just create custom functions like:

```
const handleSelect = (item, label) => {
  select(item, label);
  // your stuff
}
```

But this way you don't yet have access to updated selectedItems list, because it's going to be updated after the next render/state update, so you'd end upt needing to reimplement the whole select function from hook. Solved this by adding an effect, that will call onChange after selectedItems changes.

```
const {...} = useAutocomplete({
  onChange: (selectedItems) => {
    // your stuff using updated selectedItems
  }
})
```

### DIFF
--- a/components/autocomplete/useAutocomplete.js
+++ b/components/autocomplete/useAutocomplete.js
@@ -1,15 +1,28 @@
-import { useState } from 'react';
+import { useState, useEffect } from 'react';
+import { noop } from 'proton-shared/lib/helpers/function';
 
-const defaultItemToLabel = (label) => label;
+const defaultLabelToItem = (label) => label;
 
 const useAutocomplete = ({
     multiple = true,
     initialSelectedItems = [],
     initialInputValue = '',
-    labelToItem = defaultItemToLabel
+    labelToItem = defaultLabelToItem,
+    onChange = noop
 } = {}) => {
+    const [initialized, setInitialized] = useState(false);
     const [selectedItems, setSelectedItems] = useState(initialSelectedItems);
     const [inputValue, changeInputValue] = useState(initialInputValue);
+
+    // We want to emit onChange when items are selected or removed,
+    // but not with the initial values on mount
+    useEffect(() => {
+        if (initialized) {
+            onChange(selectedItems);
+        } else {
+            setInitialized(true);
+        }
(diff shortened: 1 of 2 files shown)


components/autocomplete/useAutocomplete.js
```diff
@@ -1,16 +1,29 @@
-import { useState } from 'react';
+import { useState, useEffect } from 'react';
+import { noop } from 'proton-shared/lib/helpers/function';
 
-const defaultItemToLabel = (label) => label;
+const defaultLabelToItem = (label) => label;
```
dhoko: Can you add the identity function inside proton-shared ?

components/autocomplete/useAutocomplete.test.js
```diff
@@ -43,4 +43,19 @@ describe('useAutocomplete hook', () => {
...
+        const initialSelectedItems = [{ label: 'test', value: 'T' }];
+        const { result } = renderHook(() =>
+            useAutocomplete({
+                multiple: true,
+                initialSelectedItems,
+                onChange: onChangeSpy
+            })
+        );
+        act(() => result.current.select(newItem, newItem.label));
+        expect(onChangeSpy).toBeCalledWith([...initialSelectedItems, newItem]);
```
dhoko: You don't test that you don't call onChange during the init ;)

vincaslt: ah, good idea

components/autocomplete/useAutocomplete.js
```diff
@@ -1,16 +1,29 @@
...
 
 const useAutocomplete = ({
     multiple = true,
     initialSelectedItems = [],
     initialInputValue = '',
-    labelToItem = defaultItemToLabel
+    labelToItem = defaultLabelToItem,
+    onChange = noop
 } = {}) => {
+    const [initialized, setInitialized] = useState(false);
```
mmso: If you don't need to trigger a re-render it'd be better to `useRef` instead

PR #82, opened 2019-05-24
Add keys functionality

mmso:


components/dropdown/DropdownActions.js
```diff
@@ -1,44 +1,60 @@
...
 import DropdownButton from './DropdownButton';
 
-const DropdownActions = ({ list, className }) => {
+const wrapTooltip = (text, tooltip) => {
+    if (!tooltip) {
+        return text;
+    }
+    return (
+        <>
+            {text}{' '}
```
EpokK: Try to not use `{' '}`, prefer:
```suggestion
            <span className="mr0-5">{text}</span>
```

components/dropdown/DropdownActions.js
```diff
@@ -1,44 +1,60 @@
...
+            <Tooltip title={tooltip}>
```
EpokK: Why about using `<Info />` component here?

components/input/FileInput.js
```diff
@@ -1,14 +1,14 @@
-import React from 'react';
+import React, { forwardRef } from 'react';
 import PropTypes from 'prop-types';
 
-const FileInput = ({ children, id, className, ...rest }) => {
+const FileInput = forwardRef(({ children, id, className, ...rest }, ref) => {
```
EpokK: I didn't know that, it's super useful.

package.json
```diff
@@ -81,6 +81,8 @@
     "proton-shared": "github:ProtonMail/proton-shared.git#master",
     "react": "^16.8.1",
     "react-router-dom": "^4.3.1",
-    "ttag": "^1.7.0"
+    "ttag": "^1.7.0",
+    "pmcrypto": "github:ProtonMail/pmcrypto.git#semver:^6.0.0",
```
dhoko: I might be wrong but it comes with proton-pack, yeah only pm-srp is not inside. We should put them here as both are mandatory for all projects.

containers/app/ErrorBoundary.js
```diff
@@ -19,17 +19,16 @@ class ErrorBoundary extends React.Component {
...
-            );
+        if (!this.state.hasError) {
+            return this.props.children;
         }
-
-        return this.props.children;
+        const error = this.state.error ? this.state.error.toString() : undefined;
+        return (
+            <>
+                <h1>Something went wrong.</h1>
```
EpokK: Missing translation?

containers/keys/reactivateKeys/ReactivateKeysList.js
```diff
@@ -0,0 +1,126 @@
...
+                                {fingerprint}
+                            </code>,
+                            keyStatus,
+                            isUpload ? uploadColumn : null
+                        ].filter(Boolean)}
+                    />
+                );
+            });
+        })
+        .flat();
```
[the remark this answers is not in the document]
mmso: yes I did that first but too complex

containers/keys/KeysActions.js
```diff
@@ -1,61 +1,83 @@
-import PropTypes from 'prop-types';
 import React from 'react';
-import { DropdownActions } from 'react-components';
+import PropTypes from 'prop-types';
 import { c } from 'ttag';
+import { DropdownActions } from 'react-components';
 
 export const ACTIONS = {
```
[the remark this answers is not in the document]
mmso: I prefer to keep it there because it's really only private to this component and section

PR #83, opened 2019-05-27
Editable text and preformatted text components

vincaslt:
Components that I needed in admin-web-panel, but that might be useful for other projects.

Editing:
Screenshot 2019-05-27 at 16 46 48

Not editing:
Screenshot 2019-05-27 at 16 46 58

Preformatted:
Screenshot 2019-05-27 at 16 46 34

### DIFF
--- a/components/editableText/EditableText.js
+++ b/components/editableText/EditableText.js
@@ -1,0 +1,84 @@
+import React, { useState, useEffect } from 'react';
+import PropTypes from 'prop-types';
+import { SmallButton, Icon, Input } from 'react-components';
+import { c } from 'ttag';
+
+const EditableText = ({ icon, onSubmit, initialText, children, readOnly, className, ...rest }) => {
+    const [inputValue, setInputValue] = useState(initialText);
+    const [editing, setEditing] = useState(false);
+
+    useEffect(() => {
+        if (editing) {
+            setInputValue(initialText);
+        }
+    }, [editing, initialText]);
+
+    const onToggleEditing = () => {
+        setEditing(!editing);
+    };
+
+    const submit = (value) => {
+        onSubmit(value);
+        setEditing(false);
+    };
+
+    const handleFormSubmit = (e) => {
+        e.preventDefault();
+        submit(inputValue);
+    };
+
+    const handleChangeInputValue = ({ target }) => setInputValue(target.value);
+
+    return editing ? (
+        <form className="flex" onSubmit={handleFormSubmit}>
+            <SmallButton onClick={onToggleEditing} className="mr0-5" title={c('Action').t`Close`}>
+                <Icon name="close" />
+            </SmallButton>
+            {children ? (
+                children({ submit })
+            ) : (
+                <>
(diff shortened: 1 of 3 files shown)


EpokK:
- Move the edit button on the right, after the value

components/editableText/EditableText.js
```diff
@@ -0,0 +1,84 @@
...
+    return editing ? (
+        <form className="flex" onSubmit={handleFormSubmit}>
+            <SmallButton onClick={onToggleEditing} className="mr0-5" title={c('Action').t`Close`}>
+                <Icon name="close" />
+            </SmallButton>
+            {children ? (
+                children({ submit })
+            ) : (
+                <>
+                    <SmallButton type="submit" className="mr0-5" title={c('Action').t`Confirm`}>
```
EpokK: You probably need to use `pm-button--for-icon` class, [see here](https://design-system-beta.netlify.com/buttons/).

components/text/Preformatted.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const Preformatted = ({ className, ...rest }) => {
+    return <pre style={{ overflow: 'auto' }} className={`bg-global-muted p0-5 m0 ${className}`} {...rest} />;
```
EpokK: Ask @nico3333fr for a class to use `overflow: 'auto'` please.

EpokK: You should have at least `mb1` on this component since it has a block structure.

components/editableText/EditableText.js
```diff
@@ -0,0 +1,84 @@
...
+        setEditing(!editing);
+    };
+
+    const submit = (value) => {
+        onSubmit(value);
+        setEditing(false);
+    };
+
+    const handleFormSubmit = (e) => {
+        e.preventDefault();
```
EpokK: Is it really needed?

components/editableText/EditableText.js
```diff
@@ -0,0 +1,84 @@
+import React, { useState, useEffect } from 'react';
+import PropTypes from 'prop-types';
+import { SmallButton, Icon, Input } from 'react-components';
+import { c } from 'ttag';
+
+const EditableText = ({ icon, onSubmit, initialText, children, readOnly, className, ...rest }) => {
+    const [inputValue, setInputValue] = useState(initialText);
+    const [editing, setEditing] = useState(false);
```
EpokK: Use `useToggle` hook and then you can use `toggle` function.

components/text/Preformatted.js
nico3333fr: You may use `scroll-if-needed` to have `overflow: auto` ;)

components/editableText/EditableText.js
vincaslt: This prevents form submission from happening. Essentially, it's not needed, because it would be prevented anyway, but with a warning thrown to console: `Form submission canceled because the form is not connected`

PR #85, opened 2019-05-28
Move string translations inside components

### LINKED ISSUE #77: Move translation inside the component
We can't define translated string outside of the component. It needs to be inside to support localization switching.

This is **BAD**

``` jsx
const I18N = {
   cheese: c('Food').t`Cheese`
};

const MyComponent = ({ action }) => {
   return I18N.cheese;
};

MyComponent.defaultProps = {
    action: c('Action').t`Jump`
};
```

This is **BETTER**

``` jsx
const MyComponent = ({ action = c('Action').t`Jump` }) => {
   const I18N = {
      cheese: c('Food').t`Cheese`
   };
   return I18N.cheese;
};
```

econdepe:
I checked all cases where `t'string to be translated'` appears in this repo. Most of them are inside components already. I moved all the rest. Hopefully I didn't miss any case.

I didn't touch the files inside `containers/keys` since @mmso was working on them

Closes #77

### DIFF
--- a/containers/invoices/InvoiceState.js
+++ b/containers/invoices/InvoiceState.js
@@ -11,15 +11,16 @@ const TYPES = {
     [INVOICE_STATE.BILLED]: 'default'
 };
 
-const InvoiceState = ({ invoice }) => {
-    const STATES = {
-        [INVOICE_STATE.UNPAID]: c('Invoice state display as badge').t`Unpaid`,
-        [INVOICE_STATE.PAID]: c('Invoice state display as badge').t`Paid`,
-        [INVOICE_STATE.VOID]: c('Invoice state display as badge').t`Void`,
-        [INVOICE_STATE.BILLED]: c('Invoice state display as badge').t`Billed`
-    };
+const getStatesI18N = () => ({
+    [INVOICE_STATE.UNPAID]: c('Invoice state display as badge').t`Unpaid`,
+    [INVOICE_STATE.PAID]: c('Invoice state display as badge').t`Paid`,
+    [INVOICE_STATE.VOID]: c('Invoice state display as badge').t`Void`,
+    [INVOICE_STATE.BILLED]: c('Invoice state display as badge').t`Billed`
+});
 
-    return <Badge type={TYPES[invoice.State]}>{STATES[invoice.State]}</Badge>;
+const InvoiceState = ({ invoice }) => {
+    const i18n = getStatesI18N();
+    return <Badge type={TYPES[invoice.State]}>{i18n[invoice.State]}</Badge>;
 };
 
 InvoiceState.propTypes = {
--- a/containers/invoices/InvoiceState.js
+++ b/containers/invoices/InvoiceState.js
@@ -4,13 +4,6 @@ import { c } from 'ttag';
 import PropTypes from 'prop-types';
 import { INVOICE_STATE } from 'proton-shared/lib/constants';
 
-const STATES = {
-    [INVOICE_STATE.UNPAID]: c('Invoice state display as badge').t`Unpaid`,
-    [INVOICE_STATE.PAID]: c('Invoice state display as badge').t`Paid`,
-    [INVOICE_STATE.VOID]: c('Invoice state display as badge').t`Void`,
-    [INVOICE_STATE.BILLED]: c('Invoice state display as badge').t`Billed`
-};
-
 const TYPES = {
     [INVOICE_STATE.UNPAID]: 'error',
     [INVOICE_STATE.PAID]: 'success',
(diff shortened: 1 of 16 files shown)


containers/login/LoginContainer.js
```diff
@@ -9,16 +9,6 @@ import TOTPForm from './TOTPForm';
 import UnlockForm from './UnlockForm';
 import useLogin from './useLogin';
 
-const getErrorText = (error) => {
-    if (error.name === 'PasswordError') {
-        return c('Error').t`Incorrect decryption password`;
```
mmso: This one is ok since it's in a function

containers/payments/subscription/CustomProfessionalSection.js
```diff
@@ -7,20 +7,18 @@ import { range } from 'proton-shared/lib/helpers/array';
...
-const domainOptions = range(2, 101).map((value, index) => ({
-    text: getTextOption('domain', value, index),
-    value: index
-}));
-
 const CustomProfessionalSection = ({ plans, model, onChange }) => {
     const professionalPlan = getPlan(plans, { name: 'professional' });
     const memberAddon = getAddon(plans, { name: '1member' });
     const domainAddon = getAddon(plans, { name: '1domain' });
+    const memberOptions = range(1, 5001).map((value, index) => ({
```
mmso: Not related to your PR but @EpokK should this be a select with 5000 options?

containers/invoices/InvoiceState.js
```diff
@@ -4,13 +4,6 @@ import { c } from 'ttag';
 import PropTypes from 'prop-types';
 import { INVOICE_STATE } from 'proton-shared/lib/constants';
 
-const STATES = {
-    [INVOICE_STATE.UNPAID]: c('Invoice state display as badge').t`Unpaid`,
-    [INVOICE_STATE.PAID]: c('Invoice state display as badge').t`Paid`,
-    [INVOICE_STATE.VOID]: c('Invoice state display as badge').t`Void`,
-    [INVOICE_STATE.BILLED]: c('Invoice state display as badge').t`Billed`
-};
```
mmso: I think I prefer this pattern:

```js
const getStatesI18N = () => ({
    [INVOICE_STATE.UNPAID]: c('Invoice state display as badge').t`Unpaid`,
    [INVOICE_STATE.PAID]: c('Invoice state display as badge').t`Paid`,
    [INVOICE_STATE.VOID]: c('Invoice state display as badge').t`Void`,
    [INVOICE_STATE.BILLED]: c('Invoice state display as badge').t`Billed`
});

const InvoiceState = ({ invoice }) => {
   const i18n = getStatesI18N();
   return <Badge type={i18n[invoice.State]}>{STATES[invoice.State]}</Badge>;
};
```

containers/login/LoginContainer.js
econdepe: Don't know why I moved it inside... :man_shrugging:

containers/invoices/InvoiceState.js
econdepe: A similar solution (defining the translation strings outside but generating them via a function inside the component) could be used in some other files I retouched. Do you want to use it in other places too (there are a few, but see `LogsTable` or `LogsSection` for examples)?

containers/logs/LogsSection.js
```diff
@@ -25,16 +25,16 @@ import { ELEMENTS_PER_PAGE, LOGS_STATE } from 'proton-shared/lib/constants';
...
-    1: c('Log event').t`Login success`,
-    2: c('Log event').t`Logout`,
-    3: c('Log event').t`2FA login failure`
-};
-
 const { DISABLE, BASIC, ADVANCED } = LOGS_STATE;
 
 const LogsSection = () => {
+    const EVENTS = {
+        0: c('Log event').t`Login password failure`,
```
EpokK: Can we also use constants here please.

containers/payments/subscription/CustomProfessionalSection.js
EpokK: We need to support existing options on v3:
Capture d’écran 2019-05-28 à 16 26 12

containers/invoices/InvoiceState.js
mmso: Yes I prefer it generally

containers/payments/subscription/CustomProfessionalSection.js
dhoko: input number ? here if the step is 1 it's... "sympa"

containers/logs/LogsSection.js
econdepe: Yeah, actually those constants are already defined. It seems there is an extra event, LOGIN_SUCCESS_AWAIT_2FA. I created a description of the event
"Login password success, awaiting 2FA"
but I'm not sure that's what it means.

containers/invoices/InvoiceState.js
econdepe: Done! It looks more homogeneous now

Issue #67, opened 2019-05-09
Add focus trap to modals


mmso:
For A11Y

mmso:
There is also this RFC in react https://github.com/reactjs/rfcs/pull/109

theKashey:
This is the most simple trap focus - https://github.com/mui-org/material-ui/blob/master/packages/material-ui/src/Modal/TrapFocus.js
This is the most complex - https://github.com/theKashey/react-focus-lock

Pick any :)

nico3333fr:
If it is possible,, when a modal is opened, if the focus trap can trap even when you tabulate from the URL also (some don't manage this), this will be perfect.

Basically, the spec is: when there is a modal, you're trapped forever with the keyboard :D

PR #90, opened 2019-05-29
Fix keyboard focus for modals and buttons

dhoko:
Most of the time when we open a modal the autofocus is not there. So we can't access to the  modal via the keyboard.
Also buttons inside the modals were broken with the keyboard.

- Add prop autoFocusClose (_default false_ to autofocus on the close button -> ex confirmModal, keyboard modal)
- Allow keyboard on enter/space for button type=reset|submit
- Autofocus Name input for the label modal

> _Why keep the preventDefault on button ?_

Because if you remove it, when you open a modal with `autoFocusClose=true` it will auto-close the modal (╯°□°）╯︵ ┻━┻ 
---> There should be a better way.

Anyway, we need to fix the focus in every modals

- [ ] tests

### DIFF
--- a/components/button/Button.js
+++ b/components/button/Button.js
@@ -33,7 +33,9 @@ const Button = ({
         }
 
         if (event.target === event.currentTarget && (key === 'space' || key === 'enter')) {
-            event.preventDefault();
+            if (!['reset', 'submit'].includes(event.target.type)) {
+                event.preventDefault();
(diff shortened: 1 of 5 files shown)


mmso:
This one is related to https://github.com/ProtonMail/react-components/issues/67 where we need to find a general solution to handle focus for the modals

PR #92, opened 2019-05-31
Add change/switch password mode modal

mmso:


containers/account/ChangePasswordModal.js
```diff
@@ -0,0 +1,394 @@
...
+                            placeholder={c('Placeholder').t`Two factor code`}
+                            required
+                        />
+                    </Field>
+                </Row>
+            )}
+            <Row>
+                <Label>{labels.newPassword}</Label>
+                <Field>
+                    <PasswordInput
```
EpokK: Since the user is supposed to enter a new password, `autocomplete` attribute should be set to `off`. Maybe we need to do something also on the `<form>`.

containers/account/ChangePasswordModal.js
```diff
@@ -0,0 +1,394 @@
...
+    };
+
+    const checkLoginError = ({ data: { Code, Error } = {} }) => {
+        if (Code === PASSWORD_WRONG_ERROR) {
+            setErrors({ loginError: Error });
+        }
+    };
+
+    const checkFatalError = (e) => {
+        if (e.name === 'NoDecryptedKeys') {
```
EpokK: This kind of error should be define in a constant, right?

PR #105, opened 2019-07-01
Remove icon button and add icon prop to standard button

vincaslt:
Added icon prop to button itself of separate IconButton. It enables better usage with Primary/Warning/Error buttons.


containers/payments/BillingSection.js
```diff
@@ -173,7 +173,7 @@ const BillingSection = ({ permission }) => {
                             </div>
                         </div>
                     ) : null}
-                    {hasPaidVpn ? (
```
vincaslt: maybe my user was corrupted somehow, but It crashed because vpnPlan was undefined, but hasPaidVpn=4

PR #107, opened 2019-07-12
colorPicker accepts hex too

econdepe:


econdepe:
Richard did it already

PR #110, opened 2019-07-17
Add separate callback for modal reset when pressing cancel button

vincaslt:
Sometimes it's useful to catch a close-by-cancel action, but not close-by-X.


mmso:
You could also handle that in the onClose callback no?

PR #113, opened 2019-07-18
Orderable table

Yiin:
Adds OrderableTable components for https://github.com/ProtonMail/proton-contacts/issues/19

Some notes:
* Components in `/components/orderable/*` are `react-sortable-hoc` wrappers.
* `OrderableTableBody` exists for consistency (it exports TableBody).
* `OrderableTable` accepts same props as Table (from `react-components`) & SortableContainer (from `react-sortable-hoc`)
* `OrderableTableRow` accepts same props as TableRow (from `react-components`) & SortableElement (from `react-sortable-hoc`)

Usage:

    <OrderableTable onSortEnd={ handleSortEnd }>
        <OrderableTableHeader cells={[ ... ]} />
        <OrderableTableBody>
            <OrderableTableRow index={ 0 } cells={[ ... ]} />
            <OrderableTableRow index={ 1 } cells={[ ... ]} />
            <OrderableTableRow index={ 2 } cells={[ ... ]} />
        </OrderableTableBody>
    </OrderableTable>


components/orderableTable/OrderableTable.scss
```diff
@@ -0,0 +1,18 @@
+.orderableHelper {
+    z-index: 666;
```
EpokK: 😈

components/orderableTable/OrderableTable.scss
```diff
@@ -0,0 +1,18 @@
...
+    display: flex;
+    align-items: center;
+    color: black;
+    background: white;
+    border-top: 1px solid #dde6ec;
+    border-bottom: 1px solid #dde6ec;
+
+    td {
```
EpokK: As best practice we try to never do CSS nesting, could you add a class on this `td`?

components/orderableTable/OrderableTable.scss
Yiin: 😄 it just so happened that 665 is z-index of the modal.

helpers/arrayMove.js
```diff
@@ -0,0 +1,22 @@
...
+};
+
+/**
+ * Returns a new array with the item moved to the new position.
+ * @param {Array} array List of items
+ * @param {number} from Index of item to move. If negative, it will begin that many elements from the end.
+ * @param {number} to Index of where to move the item. If negative, it will begin that many elements from the end.
+ * @return {Array} New array with the item moved to the new position
+ */
+export const arrayMove = (array, from, to) => {
```
mmso: These helpers would go better in proton-shared/lib/helpers/array

Also is the mutation really necessary? Did you run into performance problems?

Yiin: Good point! proton-shared/lib/helpers/array do actually have a `move` function already.

components/orderableTable/OrderableTable.scss
Yiin: So I looked more into it, and the thing is that element with `orderableHelper` class is cloned node of element that user clicked on (see: https://github.com/clauderic/react-sortable-hoc/blob/master/src/SortableContainer/index.js#L299).

Because we're not rendering that element by ourselves, we can't add these classes w/o some hacks (e.g. `[...document.querySelector('.orderableHelper td')].forEach(node => node.classList.add('column'))`

EpokK: I see what you mean, then I'm OK to keep what you did.

Issue #130, opened 2019-07-22
Handle human verification 


EpokK:
Handle human verification for all routes with captcha.

### Flow

- Frontend submits a request to Backend
- Backend detects potential abuse depending on various (route specific) parameters, responds with a dedicated response code requesting captcha (9001)
- Frontend recognises code, calls GET /users/human route to know which verification methods are available then opens a captcha verification modal, waits for response from captcha service
- Frontend resubmits the exact same request, attaches a “Token” argument on top of the original request and also TokenType=”captcha” in case we when to have different verification method in the future.
- Backend verifies Token against captcha service, either passes or denies the request

### Note

- Some logic has already be done in `feat/human-verification` branch.

PR #132, opened 2019-07-23
Add human verification support on all routes

EpokK:
### TODO

- [x] Connect `HumanVerificationModal` to 9001 handler (the error result returns the `Token` and `VerifyMethods`)
- [x] Captcha view
- [x] Modal

### Note

- The API is not yet ready.

Fix https://github.com/ProtonMail/react-components/issues/130

### DIFF
--- a/containers/api/Captcha.js
+++ b/containers/api/Captcha.js
@@ -1,0 +1,56 @@
+import React, { useState, useEffect } from 'react';
+import PropTypes from 'prop-types';
+import { useConfig } from 'react-components';
+import { getHost } from 'proton-shared/lib/helpers/url';
+import { createUrl } from 'proton-shared/lib/fetch/helpers';
+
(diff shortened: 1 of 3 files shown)


containers/api/Captcha.js
```diff
@@ -0,0 +1,56 @@
...
+
+const Captcha = ({ token, onSubmit }) => {
+    const [style, setStyle] = useState();
+    const { API_URL } = useConfig();
+    const client = 'web';
+    const host = getHost(API_URL);
+    const url = createUrl('https://secure.protonmail.com/captcha/captcha.html', { token, client, host });
+    const src = url.toString();
+
+    const receiveMessage = (event) => {
```
mmso:
```suggestion
    const handleMessage = (event) => {
```

for consistency with our other handlers?

PR #142, opened 2019-07-26
Add Image component

Yiin:
Ref: https://github.com/ProtonMail/proton-contacts/issues/28

image


components/image/Image.js
```diff
@@ -0,0 +1,19 @@
+import React, { useState } from 'react';
+import { c } from 'ttag';
+import { SHOW_IMAGES } from 'proton-shared/lib/constants';
+import Button from '../button/Button';
+import useMailSettings from '../../hooks/useMailSettings';
+
+const Image = (props) => {
```
EpokK: Since this component used `useMailSettings` hook, it should not be defined as generic component.

components/image/Image.js
```diff
@@ -0,0 +1,19 @@
...
+    const [{ ShowImages }] = useMailSettings();
```
EpokK: ⚠️ `MailSettings` are not loaded everywhere. You need to handle `loading`, the second parameter.

components/image/Image.js
```diff
@@ -0,0 +1,19 @@
...
+    const [showAnyways, setShowAnyways] = useState(false);
+
+    const handleClick = () => setShowAnyways(true);
+
+    if (ShowImages & SHOW_IMAGES.REMOTE || showAnyways) {
```
EpokK: Before to enter in this condition you need to make sure `src` is defined and is a valid URL (and not base64).

Issue #135, opened 2019-07-25
Tooltip in JS


EpokK:
We currently manage Tooltip with CSS but it doesn't work very well. 

<img width="615" alt="Capture d’écran 2019-07-25 à 16 08 50" src="https://user-images.githubusercontent.com/1345786/61881208-8484e300-aef6-11e9-85d0-c86253943c76.png">

If we cannot fix it with CSS, we need to think our proper solution in JS.

- Since the `Tooltip` is used in different places, it needs to keep the same signature.
- Managing automatic is a plus but not a requirement for the first version.

mmso:
![Screenshot 2019-07-26 at 10 15 35](https://user-images.githubusercontent.com/352607/61937078-594dd280-af8e-11e9-9053-8e179ea535c0.png)

Same problem with dropdown so let's fix that too

PR #143, opened 2019-07-26
Place tooltip over all content and adjust it's placement when it would go out of screen

vincaslt:
Fixes: #135 
Related PR: https://github.com/ProtonMail/design-system/pull/172

Images speak more than a thousand words:

hHrSRoBfmc
FOn1vkB5Gb

Some considerations:

- Right now, I only look for overflow on viewport, not some specific elements, but read the below consideration for what I think the solution could be.
- We use our own container instead of body to scroll when it's overflown, thus I need to bind scroll listener to it, and not body. I pass it as optional parameter now, or take the one with class `main`, and fallback to body when it's not passed. I think it should cover our usecases, but ideally it shouldn't be a parameter. If the project were to use another kind of container, all tooltips would need to have this class passed as params, which is cumbersome. Probably ideally, the container would be wrapped in some function, which would tell all tooltips - here's your scrollable container, the same kind of function could be used to detect overflow on elements, instead of window.


components/tooltip/utils.js
```diff
@@ -0,0 +1,39 @@
...
+};
+
+export const calculateAdjustedPosition = (target, tooltip, placement) => {
+    const reorientTo = (orientation) => calculatePosition(target, tooltip, orientation);
+
+    const originalPosition = calculatePosition(target, tooltip, placement);
+
+    if (originalPosition.top + tooltip.height > window.innerHeight) {
+        return reorientTo('top');
+    } else if (originalPosition.top < 0) {
```
EpokK: no `else if` required.

components/tooltip/Tooltip.js
```diff
@@ -1,36 +1,73 @@
...
+
+        updatePosition();
+
+        if (visible) {
+            const contentArea = document.getElementsByClassName(scrollContainerClass)[0] || document.body;
+            contentArea.addEventListener('scroll', updatePosition);
+            return () => contentArea.removeEventListener('scroll', updatePosition);
+        }
+    }, [visible, wrapperRef.current, tooltipRef.current]);
+
```
EpokK: @vincaslt @mmso do you think we need to hide the tooltip when the user change the URL?

mmso: No the element should get removed automatically

PR #145, opened 2019-07-29
Dropdown placement using JS

vincaslt:
Related PR: https://github.com/ProtonMail/design-system/pull/174

- shared logic moved to: `usePopper` hook and `Popper` component.
- reused and extended logic to support dropdown needs.
- will try to smartly recalculate orientation and position when it goes out of screen, i.e if it was at the top, it will try to stay at the top, otherwise will go to the bottom, or left/right, and vice versa.
- Added `SimpleDropdown` which basically keeps the same functionality as previous Dropdown component.
- Separated logic into reusable components, for example Dropdown could be used to show dropdown on another component than the button.
- Added additional placements for dropdowns and tooltips, now there are 8 of them for each.
- Technically, could be extended to allow limiting orientations to like 4, instead of 8.

~Right-to-left most likely doesn't work now with dropdowns...~

~RTL could be handled by detecting if layout is set to right-to-left (probz class or attribute on body or smth) and having something like this before dropdown calculations:~

```
const invertedPlacement =
    {
        'bottom-left': 'bottom-right',
        'bottom-right': 'bottom-left',
        'top-left': 'top-right',
        'top-right': 'top-left',
        left: 'right',
        right: 'left'
    }[originalPlacement] || originalPlacement;
```

DONE, using context.

As for the examples, I might add them to storybook, but for now, the usage of individual components should be clear from: https://github.com/ProtonMail/react-components/pull/145/files#diff-e3d22f8e37c4b7f34675a98da65bef04R8

For advanced use cases, basically just use:
`const { anchorRef, isOpen, toggle, close } = usePopperAnchor();`
attach `anchorRef` to element, that triggers the dropdown, and pass it to `<Dropdown anchorRef={anchorRef} />` 

`Popper` could be used to show whatever content that needs to be positioned relative to some other component and uses the same placement options, in this case it covers Tooltips and Dropdowns.


mmso:
I like it, but I would like to see even more radical changes.

Since this removes the requirement of keeping the elements tied together, I'd like to see more individual components. Where usage could be:

```js
const DropdownButton = React.forwardRef(({ onClick, open }, ref) => {
    return (
        <Button
            aria-expanded={open}
            ref={ref}
            onClick={handleToggle}
        >
            Toggle Menu Grow
            <DropdownCaret open={open}/>
        </Button>
    )
});

const MyComponentDefault = () => {
    const anchorRef = useRef();
    const [open, setOpen] = useState();

    const handleToggle = () => {
        setOpen(!open);
    };

    useClickAway(handleToggle);

    return (
        <>
            <DropdownButton ref={anchorRef} onClick={handleToggle} open={open}>
                Toggle Menu Grow
            </DropdownButton>
            <Dropdown anchor={anchorRef.current} align="right" open={open}>
                My dropdown
            </Dropdown>
        </>
    );
};
```

This would allow more freedom in the component opening the dropdown.
```js
const MyComponentAny = () => {
    const anchorRef = useRef();
    const [open, setOpen] = useState();

    const handleToggle = () => {
        setOpen(!open);
    };

    useClickAway(handleToggle);

    return (
        <>
            <th
                ref={anchorRef}
                aria-expanded={open}
                onClick={handleToggle}
            >
                Toggle Menu Grow
                <DropdownCaret open={open}/>
            </th>
            <Dropdown anchor={anchorRef.current} align="right" open={open}>
                My dropdown
            </Dropdown>
        </>
    );
};
```

What do you think about this?

Perhaps even the `useState` and `useRef` could be moved into its own hook to make it easier.

vincaslt:
I was going to add at least a component that abstracts away the portal usage. I wasn't sure how much change you'd be willing to accept, especially in the API. Some of the changes I am considering, but wasn't sure about:

- Add a wrapper abstracting away portal usage.
- Uncouple the components, because as you say, they don't need to go together anymore.
- Remove the imperative handler stuff, and accept changes to `visible` from props.
- Make tooltip and dropdown positioning consistent. Right now the tooltip has 4 different placements, dropdown has 3 (or 6 with this PR). I'd like to have 8 for both, and make the positioning more granular, rather than "flip to the left, when there is no space at the right".

Regarding the code examples you gave, I think the idea is good, and is certainly something we have to do one way or the other. I like the "slots" pattern (https://daveceddia.com/pluggable-slots-in-react-components/) for this kind of stuff though, instead of rendering components separately and linking with references.

mmso:
> How much changes

If we find an API that's better let's definitely change. The `Dropdown` component refactor weight is not that large right now.

> Add a wrapper abstracting away portal usage.

Sounds like a good idea.

> Uncouple the components, because as you say, they don't need to go together anymore.

Yes please.

> Remove the imperative handler stuff, and accept changes to visible from props.

Let's do it.

> Make tooltip and dropdown positioning consistent. 

Yes please.

> Slots pattern

I like the slots pattern too. However for this case I'm not convinced it's needed? Also how would it work with getting a ref to the anchor slot for example?

nico3333fr:
> Right now the tooltip has 4 different placements, dropdown has 3 (or 6 with this PR). I'd like to have 8 for both, and make the positioning more granular, rather than "flip to the left, when there is no space at the right".

nico3333fr:
> Right now the tooltip has 4 different placements, dropdown has 3 (or 6 with this PR). I'd like to have 8 for both, and make the positioning more granular, rather than "flip to the left, when there is no space at the right".

I may do positionning using CSS, and JS may decide which one is adapted, or overpass standard positionning.

components/dropdown/Dropdown.js
```diff
@@ -1,133 +1,92 @@
...
-        ref
-    ) => {
-        const [open, setOpen] = useState(isOpen);
-        const wrapperRef = useRef(null);
-        const handleClick = () => setOpen(!open);
+const Dropdown = ({
+    anchorRef,
+    children,
+    originalPlacement = 'bottom',
+    close = noop,
```
mmso: I'd prefer to call this `onClose` instead of `close` for consistency with other handlers, what do you think?

vincaslt: yep, good point

components/Popper/usePopper.js
```diff
@@ -0,0 +1,61 @@
...
+    popperRef,
+    anchorRef,
+    visible,
+    { originalPlacement = 'bottom', offset = 10, scrollContainerClass = null } = {}
+) => {
+    const [placement, setPlacement] = useState(originalPlacement);
+    const [position, setPosition] = useState({ top: -1000, left: -1000 });
+
+    useEffect(() => {
+        const updatePosition = () => {
```
EpokK: Is it necessary to put this function in `useEffect`? I prefer to move it outside if possible.

components/tooltip/usePopper.js
```diff
@@ -0,0 +1,67 @@
...
+ * Available placements: bottom-left, bottom, bottom-right, top-left, top, top-right, left, right
+ */
+const usePopper = ({ placement = 'bottom', visible: visibleInitially = false, scrollContainerClass, offset }) => {
+    const tooltipRef = useRef();
+    const wrapperRef = useRef();
+    const [visible, setVisible] = useState(visibleInitially);
+    const [position, setPosition] = useState({ top: -1000, left: -1000, placement });
+
+    useEffect(() => {
+        const updatePosition = () => {
```
EpokK: Same feedback.

components/dropdown/Dropdown.js
```diff
@@ -1,133 +1,97 @@
-import React, { useState, useEffect, useRef, useImperativeHandle } from 'react';
+import React, { useEffect, useRef } from 'react';
 import PropTypes from 'prop-types';
 import keycode from 'keycode';
-import { Icon } from 'react-components';
+import { classnames } from '../../helpers/component';
+import { usePopper, Popper } from '../Popper';
+import { noop } from '@babel/types';
```
mmso:
```suggestion
import { noop } from 'proton-shared/lib/helpers/function';
```

containers/error/GenericError.js
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 import { c } from 'ttag';
 import { PropTypes, IllustrationPlaceholder } from 'prop-types';
 
-import errorImg from 'design-system/assets/img/shared/generic-error.svg';
+// import errorImg from 'design-system/assets/img/shared/generic-error.svg';
```
mmso: It exists now

components/dropdown/Dropdown.js
vincaslt: autoimport fail

containers/error/GenericError.js
vincaslt: accidental commit :D

components/Popper/usePopper.js
mmso: I prefer to keep it inside because then it makes it clearer when it's run and the function is only created when it's needed

vincaslt: if it was outside, I would need to wrap it in useCallback, which makes it uglier :D

PR #151, opened 2019-07-30
Disable sticky labels when not in conversation mode

Yiin:
Fix: https://github.com/ProtonMail/protonmail-settings/issues/70

### DIFF
--- a/containers/layouts/LayoutsSection.js
+++ b/containers/layouts/LayoutsSection.js
@@ -21,7 +21,7 @@ import {
     updateDraftType,
     updateRightToLeft
 } from 'proton-shared/lib/api/mailSettings';
-import { VIEW_MODE } from 'proton-shared/lib/constants';
+import { VIEW_MODE, STICKY_LABELS } from 'proton-shared/lib/constants';
 
 import DraftTypeSelect from './DraftTypeSelect';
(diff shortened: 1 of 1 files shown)


containers/layouts/LayoutsSection.js
```diff
@@ -55,6 +55,9 @@ const LayoutsSection = () => {
     };
 
     const handleChangeViewMode = async (mode) => {
+        if (mode !== VIEW_MODE.GROUP) {
```
EpokK: Why not using a truly condition?

containers/layouts/LayoutsSection.js
```diff
@@ -55,6 +55,9 @@ const LayoutsSection = () => {
...
+            await api(updateStickyLabels(0));
```
EpokK: Use a constant please. If it doesn't exist, create it in `proton-shared/lib/constants`.

containers/layouts/LayoutsSection.js
Yiin: This is my reasoning that I thought about for a second before writing the code:
The logic as I understand is, sticky labels can only be enabled when view mode is set to group, so if we add new view mode that also does not support sticky labels, instead of adding OR clause we wouldn't need to change anything. I can change it to truthy condition if that's not a case.

Issue #119, opened 2019-07-18
Close modals when navigating


mmso:
When the user is navigating to a new route, we need to close the modal.

EpokK:
Yes we need to.

PR #157, opened 2019-08-01
Reset modals when navigating

vincaslt:
Fixes: #119 
Also clean some console.errors


mmso:
This is the approach I was thinking of doing too. But because we can have promises waiting for modals to be resolved I think I'd prefer to call the onClose callback on each of them. That would also have the benefit of animating them out. What do you think? 

```js
useEffect(() => {
       latestModals.current.forEach((modal) => {
             content.props.onClose && content.props.onClose();
             hideModal(id);
       });
}, [location]);
```

vincaslt:
Great point about the promises. Will change.

PR #162, opened 2019-08-05
Fix filter condition editing and deleting

vincaslt:
Fixes: https://github.com/ProtonMail/protonmail-settings/issues/88
And additional bug I noticed, where filter condition would not be saved after editing (only local state was updated, not the model).

### DIFF
--- a/containers/filters/editor/EditConditionValue.js
+++ b/containers/filters/editor/EditConditionValue.js
@@ -1,24 +1,16 @@
-import React, { useState } from 'react';
+import React from 'react';
 import PropTypes from 'prop-types';
 import { c } from 'ttag';
 import { Input, Button } from 'react-components';
 import { noop } from 'proton-shared/lib/helpers/function';
 
 function EditConditionValue({ value, onClickDelete, onEdit }) {
-    const [state, setState] = useState(value);
-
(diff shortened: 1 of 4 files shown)


containers/filters/editor/EditConditionValue.js
```diff
@@ -1,24 +1,16 @@
...
 import { c } from 'ttag';
 import { Input, Button } from 'react-components';
 import { noop } from 'proton-shared/lib/helpers/function';
 
 function EditConditionValue({ value, onClickDelete, onEdit }) {
-    const [state, setState] = useState(value);
-
     const handleClick = () => onClickDelete(value);
-    const handleInput = ({ target }) => setState(target.value);
-    const handleKeyUp = ({ key }) => {
```
EpokK: Where did you move this logic?

vincaslt: When editing, there is an input field with a button `Delete`, in contrast to input+`Add` when creating filters. When creating filter, pressing enter adds a condition, but when editing, it submits the whole form.

Not sure about what the behavior should be, but I removed the onKeyUp, as it did nothing in "edit" case. Now the model value updates on every change in the input field.

PR #164, opened 2019-08-05
Add color to plan in main Logo

econdepe:
Use new classes from design-system (https://design-system-beta.netlify.com/colors/)

### DIFF
--- a/components/logo/MailLogo.js
+++ b/components/logo/MailLogo.js
@@ -3,8 +3,6 @@ import PropTypes from 'prop-types';
 
 import { PLAN_NAMES } from 'proton-shared/lib/constants';
 
-const planNames = ['', ...Object.keys(PLAN_NAMES)];
-
 const MailLogo = ({ planName = '' }) => {
     return (
         <svg
@@ -49,7 +47,7 @@ const MailLogo = ({ planName = '' }) => {
 };
 
 MailLogo.propTypes = {
-    planName: PropTypes.oneOf(planNames)
+    planName: PropTypes.oneOf(Object.keys(PLAN_NAMES))
 };
 
 export default MailLogo;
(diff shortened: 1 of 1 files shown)


components/logo/MailLogo.js
```diff
@@ -38,7 +45,7 @@ const MailLogo = ({ planName = '' }) => {
 };
 
 MailLogo.propTypes = {
-    planName: PropTypes.string
+    planName: PropTypes.oneOf(['', 'plus', 'proffesional', 'visionary'])
```
EpokK: Use constants please.

EpokK: ⚠️ you wrote `proffesional`.

components/logo/MailLogo.js
```diff
@@ -29,7 +29,14 @@ const MailLogo = ({ planName = '' }) => {
...
             <title id="logo__title">ProtonMail</title>
             {planName ? (
-                <text textAnchor="end" className="plan uppercase bold" x="147" y="42" id="plan" focusable="false">
+                <text
+                    textAnchor="end"
+                    className={`plan fill-${planName} uppercase bold`}
+                    x="147"
+                    y="42"
+                    id="plan"
+                    focusable="false"
```
EpokK:
```suggestion
                    focusable={false}
```

components/logo/MailLogo.js
econdepe: not my best day...

components/logo/MailLogo.js
```diff
@@ -1,6 +1,10 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
+import { PLAN_NAMES } from 'proton-shared/lib/constants';
+
+const planNames = ['', ...Object.keys(PLAN_NAMES)];
```
EpokK: I don't think you need to add `''` since `planName` prop is not required.

econdepe: mmhh, that's a good point!

Issue #152, opened 2019-07-31
Upgrade button


EpokK:
Create a link component "Upgrade" to put inside the header for `ProtonContacts` and `ProtonCalendar` that pop-up if the user is free (use `useUser` hook). If the user click on it, redirect him to `/settings/subscription`. 

<img width="1232" alt="Capture d’écran 2019-07-31 à 03 22 59" src="https://user-images.githubusercontent.com/1345786/62176616-83aedf80-b342-11e9-80bf-497c2c88ac22.png">

PR #166, opened 2019-08-06
Add UpgradeButton (#152)

Yiin:
The popup opens on /settings/subscription page if the user is on the free plan (PlansSection component), so we simply redirect him to that page and let it handle everything (same behavior is on Angular 4 project).

### DIFF
--- a/containers/heading/UpgradeButton.js
+++ b/containers/heading/UpgradeButton.js
@@ -1,5 +1,4 @@
 import React from 'react';
-import { Link } from 'react-router-dom';
 import PropTypes from 'prop-types';
 import { c } from 'ttag';
 
@@ -7,10 +6,10 @@ import { Icon } from 'react-components';
 
 const UpgradeButton = ({ className }) => {
     return (
-        <Link to="/settings/subscription" className={className}>
+        <a to="/settings/subscription" className={className}>
             <Icon name="upgrade-to-paid" className="topnav-icon mr0-5 flex-item-centered-vert fill-white" />
             {c('Link').t`Upgrade`}
-        </Link>
+        </a>
     );
 };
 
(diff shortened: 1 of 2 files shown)


containers/heading/UpgradeButton.js
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import { Link } from 'react-router-dom';
+import PropTypes from 'prop-types';
+import { c } from 'ttag';
+
+import { Icon } from 'react-components';
+
+const UpgradeButton = ({ className }) => {
+    return (
+        <Link to="/settings/subscription" className={className}>
```
EpokK: Just check if `Link` works on apps that doesn't support this route please.

containers/heading/UpgradeButton.js
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { c } from 'ttag';
+
+import { Icon } from 'react-components';
+
+const UpgradeButton = ({ className }) => {
+    return (
+        <a to="/settings/subscription" className={className}>
```
EpokK: ⚠️ 
```suggestion
        <a href="/settings/subscription" className={className}>
```

containers/heading/UpgradeButton.js
```diff
@@ -0,0 +1,20 @@
...
+            <Icon name="upgrade-to-paid" className="topnav-icon mr0-5 flex-item-centered-vert fill-white" />
```
EpokK: Since the `Icon` is formatted to be in the top bar, I'm wondering if it makes sense to keep `className` as a prop.

containers/heading/UpgradeButton.js
Yiin: it doesn't

containers/heading/UpgradeButton.js
Yiin: I'll update it next time I work on header navigation. 👍

Issue #165, opened 2019-08-06
Dropdowns inside modals not displaying


econdepe:
When a dropdown is inside a modal, its content is hidden behind the modal (apparently dropdown has `z-index = 10` plus a fixed position while the modal has `z-index = 666`).

I'm using the `DropdownActions` component which uses the new dropdowns. See attached pictures below:
![image](https://user-images.githubusercontent.com/33841139/62552203-8abb7d80-b86d-11e9-8687-af60095e7ed9.png)
![image](https://user-images.githubusercontent.com/33841139/62552289-b0488700-b86d-11e9-8737-a4ba9e8aae77.png)

According to @nico3333fr , one good solution would be for the dropdown to detect that it's been opened inside a modal, then set a `z-index > 666`. If setting a `z-index > 666` one could run into other unexpected problems if the dropdown is opened simultaneously with other elements, although these are unlikely situations

mmso:
We should render the modals in a portal too, so they are rendered as siblings to the drop downs and render them both at the same z-index, given that everything is inserted in order it should be ok.

PR #168, opened 2019-08-07
Portalize modals

vincaslt:
All modals will render into `.modal-root` via portals.

Related PRs to fix dropdowns and tooltips in modals:
`design-system`: https://github.com/ProtonMail/design-system/pull/185
~`proton-contacts`: https://github.com/ProtonMail/proton-contacts/pull/80~ ✅
~new one^: https://github.com/ProtonMail/proton-contacts/pull/81~ ✅

Fixes: #165


mmso:
I think we should remove the `.modal-root` container and make the `.pm-modalOverlay` have the same z-index. 

What I want to achieve is the following in the portal:

```
<body>
  <App>
    <-AppDropdown/>
  </App>
  <Dropdown (from app)/>
  <ModalOverlay/>
  <Modal/>
  <Dropdown/>
  <Another kind of overlay between modals/>
  <Modal/>
  <Dropdown/>
</body>
```

Because it solves everything naturally

mmso:
Portals didn't work as I expected, need to think more how to do it better.

mmso:
```js
const Portal = ({ container, children }) => {
    const [target, setTarget] = useState(null);
    useLayoutEffect(() => {
        setTarget(ReactDOM.findDOMNode(container) || document.body);
        return () => setTarget(null);
    }, [container]);
    return target ? ReactDOM.createPortal(children, target) : null;
};
```

Seems pretty clean

PR #169, opened 2019-08-07
Add support for PayPal recurring and 3DS

EpokK:
- Update PayPal view with the new payment token logic
- Update payment methods table
- Update payment methods select


emnproton:
The gateway page at secure.protonmail.blue/com does not auto-close once the transaction is successfully completed.

Unless the user closes the tab, they have no way of knowing that their payment went through.

**Video:**

[secure page doesnt auto close.zip](https://github.com/ProtonMail/react-components/files/3477360/secure.page.doesnt.auto.close.zip)

containers/payments/PayPal.js
```diff
@@ -1,43 +1,176 @@
...
+    const [loading, withLoading] = useLoading();
+    const tabRef = useRef(null);
+    const timerRef = useRef(0);
+    const processRef = useRef(true);
+    const tokenRef = useRef('');
+    const api = useApi();
+    const [approvalURL, setAppovalURL] = useState('');
+    const [error, setError] = useState(null);
+
+    const I18N = {
```
Yiin: It would be better to move this outside the component, to avoid repeated calls on re-renders.

EpokK: This needs to be inside the component to refresh it when the language is changed.

containers/payments/tokenHelper.js
```diff
@@ -0,0 +1,135 @@
...
+
+        const onMessage = (event) => {
+            const origin = event.origin || event.originalEvent.origin; // For Chrome, the origin property is in the event.originalEvent object.
+
+            if (origin !== 'https://secure.protonmail.com') {
+                return reject();
+            }
+
+            if (event.source !== tab) {
+                return reject();
```
mmso: I don't think we need to `reject()` here or at the origin check

EpokK: But if I don't `reject` here, who take care of the promise process? I don't like to count on `listenTab` for it.

EpokK: Oh I understand your point now. You are right.

PR #178, opened 2019-08-09
Add Tabs component

EpokK:
@nico3333fr will add the appropriate style on design-system.

### DIFF
--- a/components/container/Tabs.js
+++ b/components/container/Tabs.js
@@ -1,30 +1,29 @@
-import React, { useState } from 'react';
+import React from 'react';
 import PropTypes from 'prop-types';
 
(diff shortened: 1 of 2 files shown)


components/container/Tabs.js
```diff
@@ -0,0 +1,52 @@
+import React, { useState } from 'react';
+import PropTypes from 'prop-types';
+
+const toKey = (index, prefix = '') => `${prefix}${index}`;
+
+const Tabs = ({ panels = [], index = 0 }) => {
```
mmso: What do you think about making it a controlled component by default?

We can add another `SimpleTabs` component that keeps state.

PR #185, opened 2019-08-12
Remove wrong default prop in IllustrationPlaceholder

econdepe:
With `text=''`, if `text` was not specified, the component would display a useless `<p></p>`


emnproton:
Verified as fixed.

PR #193, opened 2019-08-13
Storage space status & CircularProgress component (#179)

### LINKED ISSUE #179: Add storage space popup in AppsSidebar
### Design

https://zpl.io/VqLNNg6

Capture d’écran 2019-08-11 à 21 32 19

### Note
- The storage pop-up appears when the user click on the icon in the left sidebar
- Upgrade goes to `/settings/subscription`

Yiin:
Fix #179 
image

### DIFF
--- a/components/dropdown/Dropdown.js
+++ b/components/dropdown/Dropdown.js
@@ -9,10 +9,13 @@ import { noop } from 'proton-shared/lib/helpers/function';
 const Dropdown = ({
     anchorRef,
     children,
+    className,
     originalPlacement = 'bottom',
     onClose = noop,
     isOpen = false,
     narrow = false,
+    wide = false,
+    auto = false,
     autoClose = true,
     autoCloseOutside = true,
     ...rest
@@ -67,7 +70,14 @@ const Dropdown = ({
         };
     }, []);
 
-    const contentClassName = classnames(['dropDown', `dropDown--${placement}`, narrow && 'dropDown--narrow']);
+    const contentClassName = classnames([
+        'dropDown',
+        `dropDown--${placement}`,
+        narrow && 'dropDown--narrow',
+        wide && 'dropDown--wide',
+        auto && 'dropDown--auto',
+        className
+    ]);
     return (
         <Popper
             ref={popperRef}
@@ -86,10 +96,13 @@ const Dropdown = ({
 Dropdown.propTypes = {
     anchorRef: PropTypes.shape({ current: PropTypes.instanceOf(Element) }),
     children: PropTypes.node.isRequired,
+    className: PropTypes.string,
     onClose: PropTypes.func,
     isOpen: PropTypes.bool,
     originalPlacement: PropTypes.string,
     narrow: PropTypes.bool,
+    wide: PropTypes.bool,
+    auto: PropTypes.bool,
     autoClose: PropTypes.bool,
     autoCloseOutside: PropTypes.bool
 };
--- a/components/dropdown/Dropdown.js
+++ b/components/dropdown/Dropdown.js
@@ -13,9 +13,7 @@ const Dropdown = ({
     originalPlacement = 'bottom',
     onClose = noop,
     isOpen = false,
-    narrow = false,
-    wide = false,
-    auto = false,
+    size = 'normal',
     autoClose = true,
     autoCloseOutside = true,
     ...rest
@@ -73,9 +71,7 @@ const Dropdown = ({
     const contentClassName = classnames([
         'dropDown',
         `dropDown--${placement}`,
-        narrow && 'dropDown--narrow',
-        wide && 'dropDown--wide',
-        auto && 'dropDown--auto',
+        size !== 'normal' && `dropDown--${size}`,
         className
     ]);
     return (
@@ -100,9 +96,7 @@ Dropdown.propTypes = {
     onClose: PropTypes.func,
     isOpen: PropTypes.bool,
     originalPlacement: PropTypes.string,
-    narrow: PropTypes.bool,
-    wide: PropTypes.bool,
-    auto: PropTypes.bool,
+    size: PropTypes.oneOf(['normal', 'narrow', 'wide', 'auto']),
     autoClose: PropTypes.bool,
     autoCloseOutside: PropTypes.bool
 };
--- a/components/progress/CircularProgress.js
+++ b/components/progress/CircularProgress.js
@@ -1,8 +1,8 @@
-import React, { useMemo } from 'react';
+import React from 'react';
 import PropTypes from 'prop-types';
+import { classnames } from '../../helpers/component';
 
 import './CircularProgress.scss';
-import { classnames } from '../../helpers/component';
 
 const CircularProgress = ({
     children,
@@ -10,24 +10,10 @@ const CircularProgress = ({
     progress,
     rootRef,
     size = 32,
-    backgroundColor = '#555',
-    color = '#4cc75a',
+    backgroundColor = 'global-altgrey',
+    color = 'pm-blue',
     ...rest
 }) => {
-    const backgroundCircleStyle = useMemo(
-        () => ({
-            stroke: backgroundColor
-        }),
-        backgroundColor
-    );
-
-    const circleStyle = useMemo(
-        () => ({
-            stroke: color
-        }),
-        color
-    );
-
     return (
         <svg
             ref={rootRef}
@@ -38,20 +24,18 @@ const CircularProgress = ({
             {...rest}
         >
             <path
-                style={backgroundCircleStyle}
-                className="circle"
+                className={classnames(['circle', `stroke-${backgroundColor}`])}
                 strokeDasharray="100, 100"
                 d="M18 2.0845
-            a 15.9155 15.9155 0 0 1 0 31.831
-            a 15.9155 15.9155 0 0 1 0 -31.831"
+                    a 15.9155 15.9155 0 0 1 0 31.831
+                    a 15.9155 15.9155 0 0 1 0 -31.831"
             />
             <path
-                circleStyle={circleStyle}
-                className="circle"
+                className={classnames(['circle', `stroke-${color}`])}
                 strokeDasharray={`${progress}, 100`}
                 d="M18 2.0845
-            a 15.9155 15.9155 0 0 1 0 31.831
-            a 15.9155 15.9155 0 0 1 0 -31.831"
+                    a 15.9155 15.9155 0 0 1 0 31.831
+                    a 15.9155 15.9155 0 0 1 0 -31.831"
             />
             {children}
         </svg>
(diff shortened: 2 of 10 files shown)


containers/app/StorageSpaceStatus.js
```diff
@@ -0,0 +1,74 @@
...
+                                <span className="centered-absolute">{usedPercent}%</span>
+                            </div>
+                        </div>
+                        <div className="w150p">
+                            <div className="big mt0 mb0">Storage</div>
+                            <div className="small color-black mt0 mb0">
+                                {usedSpaceFormatted} of {maxSpaceFormatted} used
+                            </div>
+                            <p className="smaller opacity-40 mt0 mb1">
+                                Your storage space is shared across all Proton products.
```
EpokK: Translation needed.

containers/app/StorageSpaceStatus.js
```diff
@@ -0,0 +1,74 @@
...
+                        <div className="pr1 flex flex-items-center">
+                            <div className="relative">
+                                <CircularProgress progress={usedPercent} size={100} backgroundColor="#eeeeee" />
+                                <span className="centered-absolute">{usedPercent}%</span>
+                            </div>
+                        </div>
+                        <div className="w150p">
+                            <div className="big mt0 mb0">Storage</div>
+                            <div className="small color-black mt0 mb0">
+                                {usedSpaceFormatted} of {maxSpaceFormatted} used
```
EpokK: Translation needed.

containers/app/StorageSpaceStatus.js
```diff
@@ -0,0 +1,74 @@
...
+                    </div>
+                    <div className="flex">
+                        <div className="pr1 flex flex-items-center">
+                            <div className="relative">
+                                <CircularProgress progress={usedPercent} size={100} backgroundColor="#eeeeee" />
+                                <span className="centered-absolute">{usedPercent}%</span>
+                            </div>
+                        </div>
+                        <div className="w150p">
+                            <div className="big mt0 mb0">Storage</div>
```
EpokK: Translation needed.

containers/app/AppsSidebar.js
```diff
@@ -34,6 +35,10 @@ const AppsSidebar = ({ currentApp = '' }) => {
                         </li>
                     );
                 })}
+                <li className="flex-item-fluid" />
+                <li key="storageStatus" className="mb0-5">
```
EpokK: Do you need `key` prop here?

containers/app/StorageSpaceStatus.js
```diff
@@ -0,0 +1,74 @@
...
+                </text>
+            </CircularProgress>
+            <span className="center opacity-40 smaller storage">{usedSpaceFormatted}</span>
+            <Dropdown
+                id={uid}
+                isOpen={isOpen}
+                anchorRef={anchorRef}
+                onClose={close}
+                originalPlacement="left-bottom"
+                auto
```
EpokK:
```suggestion
                auto={true}
```

components/dropdown/Dropdown.js
```diff
@@ -9,10 +9,13 @@ import { noop } from 'proton-shared/lib/helpers/function';
 const Dropdown = ({
     anchorRef,
     children,
+    className,
     originalPlacement = 'bottom',
     onClose = noop,
     isOpen = false,
     narrow = false,
+    wide = false,
```
vincaslt: not a fan of conflicting props - it's not immediately clear what the result of `wide=true` & `narrow=true` is

`size: 'narrow' | 'wide'` is probably better

components/progress/CircularProgress.js
```diff
@@ -0,0 +1,71 @@
...
+    ...rest
+}) => {
+    const backgroundCircleStyle = useMemo(
+        () => ({
+            stroke: backgroundColor
+        }),
+        backgroundColor
+    );
+
+    const circleStyle = useMemo(
```
mmso: Since this is not really an expensive computation, I think it's fine to not use `useMemo` in this case.

components/progress/CircularProgress.js
```diff
@@ -0,0 +1,71 @@
...
+    size = 32,
+    backgroundColor = '#555',
+    color = '#4cc75a',
+    ...rest
+}) => {
+    const backgroundCircleStyle = useMemo(
+        () => ({
+            stroke: backgroundColor
+        }),
+        backgroundColor
```
vincaslt: I don't think it's expected to pass non-array value as a dependency list. At least judging from typescript definitions it's:

```
type DependencyList = ReadonlyArray<any>;
function useMemo<T>(factory: () => T, deps: DependencyList | undefined): T;
```

components/progress/CircularProgress.js
```diff
@@ -0,0 +1,71 @@
...
+    );
+
+    const circleStyle = useMemo(
+        () => ({
+            stroke: color
+        }),
+        color
```
vincaslt: same

components/progress/CircularProgress.js
```diff
@@ -0,0 +1,71 @@
...
+    children,
+    className,
+    progress,
+    rootRef,
+    size = 32,
+    backgroundColor = '#555',
+    color = '#4cc75a',
+    ...rest
+}) => {
+    const backgroundCircleStyle = useMemo(
```
vincaslt: why memo though?

components/progress/CircularProgress.js
mmso: Yes it needs to be an array 👍

containers/app/StorageSpaceStatus.js
```diff
@@ -0,0 +1,74 @@
...
+                rootRef={anchorRef}
+                onClick={toggle}
+            >
+                <text
+                    x="50%"
+                    y="50%"
+                    fontFamily="Constantia"
+                    textAnchor="middle"
+                    alignmentBaseline="middle"
+                    stroke="#ffffff"
```
vincaslt: judging from the image it's supposed to be `right-bottom`. The first part says on which side of target to place the dropdown, the second one says how it's aligned in relation to the target.

Glad my calculation works though

containers/app/StorageSpaceStatus.js
```diff
@@ -0,0 +1,74 @@
...
+                    i
+                </text>
+            </CircularProgress>
+            <span className="center opacity-40 smaller storage">{usedSpaceFormatted}</span>
+            <Dropdown
+                id={uid}
+                isOpen={isOpen}
+                anchorRef={anchorRef}
+                onClose={close}
+                originalPlacement="left-bottom"
```
vincaslt: judging from the image it's supposed to be `right-bottom`. The first part says on which side of target to place the dropdown, the second one says how it's aligned in relation to the target.

Glad my calculation works though

containers/app/StorageSpaceStatus.js
```diff
@@ -0,0 +1,74 @@
...
+                <div className="dropDown-content p1">
+                    <div className="absolute top-right mt0-5 mr0-5">
+                        <button className="flex flex-items-center">
+                            <Icon name="close" />
+                        </button>
+                    </div>
+                    <div className="flex">
+                        <div className="pr1 flex flex-items-center">
+                            <div className="relative">
+                                <CircularProgress progress={usedPercent} size={100} backgroundColor="#eeeeee" />
```
vincaslt: Should we maybe include these colors in the design system and use as classes? I don't particularly care how we use them (js constants, classnames, whatever) but it would be good to keep track of them somehow.

containers/app/StorageSpaceStatus.js
```diff
@@ -0,0 +1,74 @@
...
+                        </div>
+                        <div className="w150p">
+                            <div className="big mt0 mb0">{c('Title').t`Storage`}</div>
+                            <div className="small color-black mt0 mb0">
+                                {c('Info').jt`${usedSpaceFormatted} of ${maxSpaceFormatted} used`}
+                            </div>
+                            <p className="smaller opacity-40 mt0 mb1">
+                                {c('Info').t`Your storage space is shared across all Proton products.`}
+                            </p>
+                            <a href="/settings/subscription" className="pm-button-blue">{c('Action').t`Upgrade`}</a>
```
mmso: Should use `Link` if you are in the settings app, otherwise it'd refresh the page I think?

mmso: Or perhaps each app should send a upgrade component to let each app define the behaviour themselves, since the URL will differ in the future

containers/app/StorageSpaceStatus.js
EpokK: I agree, import colors from `design-system` and inject it in the component.

containers/app/StorageSpaceStatus.js
EpokK: Can we detect if the route goes inside the app or outside by checking `Routes` with `react-router`? Because, I'm thinking of using a component for it.

containers/app/StorageSpaceStatus.js
nico3333fr: Yes, you can use `fill-*` for SVG or `bg-` classes, will be easier also to theme later :) 
https://design-system-beta.netlify.com/colors/

nico3333fr: If you need other colors, why not, but I prefer using colors defined by designers, even if this looks not perfect. 
I prefer hearing them saying "Let's improve this" instead of "DO NOT INVENT DESIGN PLZ!" :D

containers/app/StorageSpaceStatus.js
Yiin: I'm thinking about checking if the first part of the pathname matches the first part of the link.
```
    const [, app] = url.split('/');
    const [, currentApp] = location.pathname.split('/');

    if (app && currentApp === app) {
        return (
            <Link to={url} className={className} rel={rel}>
                {children}
            </Link>
        );
    }
```

components/dropdown/Dropdown.js
```diff
@@ -86,10 +92,11 @@ const Dropdown = ({
 Dropdown.propTypes = {
     anchorRef: PropTypes.shape({ current: PropTypes.instanceOf(Element) }),
     children: PropTypes.node.isRequired,
+    className: PropTypes.string,
     onClose: PropTypes.func,
     isOpen: PropTypes.bool,
     originalPlacement: PropTypes.string,
-    narrow: PropTypes.bool,
+    size: PropTypes.oneOf(['narrow', 'wide', 'auto']),
```
mmso: Should probably also allow to pass `default`, or maybe we can call it `normal`?

containers/app/StorageSpaceStatus.js
mmso: There will not be a 1-1 mapping between the pathname and the app so we need another solution

mmso: I suggest passing a `upgradeUrl` or `<Upgrade>` component from the app

Issue #174, opened 2019-08-09
TextArea with autoGrow support


EpokK:
Add an option on `TextArea` component to support auto grow effect. The input needs to auto change his size when user type things inside.

### Note

- Example: http://www.jacklmoore.com/autosize/ try to right our own logic for that, don't use a library.
- Name the prop `autoGrow={Boolean}` (default `false`)
- The component start with 1 line
- The limit of line is defined by existing `rows` prop.

PR #194, opened 2019-08-13
Autogrow textarea (#174)

Yiin:


components/input/TextArea.js
```diff
@@ -1,14 +1,54 @@
-import React, { useState } from 'react';
+import React, { useState, useCallback } from 'react';
 import PropTypes from 'prop-types';
 
 import { generateUID } from '../../helpers/component';
 import useInput from './useInput';
 import ErrorZone from '../text/ErrorZone';
 
+const useAutoGrow = ({ maxRows = 5, minRows = 1, autoGrow = false }, handlers) => {
```
EpokK: I'm fine to put `useAutoGrow` as a separate file instead of inside `TextArea`.

PR #196, opened 2019-08-13
Remove defaultProps

### LINKED ISSUE #159: Remove `Proptype...isRequired` from props that have default value specified through ES6
Some components have props with a default value specified through ES6 notation. If those props are marked as `isRequired` in `component.propTypes`, React throws a warning even though the component works well because it does take the default value specified through ES6 notation.

If the default value is specified through `defaultProps` this problem does not exist, but since we're dropping it, this behavior should be changed. At any rate it does not make much sense to require a prop that has a default value. 

Examples:
https://github.com/ProtonMail/react-components/blob/master/components/alert/Alert.js
https://github.com/ProtonMail/react-components/blob/master/components/modal/Content.js

econdepe:
Specify default props using ES6 notation instead of `defaultProps`.

Mostly trivial changes, but pay special attention to Input and Select components

Closes #159


components/input/Input.js
```diff
@@ -4,26 +4,66 @@ import PropTypes from 'prop-types';
...
+            onFocus,
+            onKeyDown,
+            onKeyUp,
+            onPressEnter,
+            placeholder,
+            readOnly,
+            required,
+            type = 'text',
+            value
+        },
```
mmso: I'd like to keep passing rest if possible, something like this:

```js
const Input = React.forwardRef(({ className = '', error, autoComplete = 'off', type = 'text', onPressEnter, ...rest }, ref) => {
    const { handlers, statusClasses, status } = useInput({ ...rest, onPressEnter });
    const [uid] = useState(generateUID('input'));
    return (
        <>
            <input
                className={`pm-field w100 ${className} ${statusClasses}`}
                aria-invalid={error && status.isDirty}
                aria-describedby={uid}
                ref={ref}
                autoComplete={autoComplete}
                type={type}
                {...rest}
                {...handlers}
            />
            <ErrorZone id={uid}>{error && status.isDirty ? error : ''}</ErrorZone>
        </>
    );
});
```
(and same for Select)

Do you see a reason that wouldn't work?

econdepe: I hesitated whether to write that, but I thought maybe it was a strange mix (some props types defined but the props not declared, some not defined nor declared, some defined and declared). But definitively it looks better code-wise, so if you prefer that, I'm all for it!

PR #200, opened 2019-08-15
WIP: Remove use of moment.js

### LINKED ISSUE #116: Unify date libraries
Right now we are using `moment`, `dayjs` and `date-fns`. Unify the usage to only use one.

econdepe:
Migrate from 'moment' library to 'date-fns'.

Requires https://github.com/ProtonMail/proton-shared/pull/44

WIP: Moving locale loading into a provider

# closes #116


econdepe:
'Pikaday' library uses 'Moment.js' for custom formatting, which we use in the components `StartDateField` and `EndDateField`. I'll try to write some sort of wrapper

containers/autoReply/AutoReplyForm/fields/StartDateField.js
```diff
@@ -18,7 +20,7 @@ const StartDateField = ({ value, onChange }) => {
                     defaultDate={new Date(value)}
                     setDefaultDate
                     onSelect={handleChange}
-                    format={moment.localeData().longDateFormat('L')}
+                    format={detectLongDateFormat(currentLocale)}
```
vincaslt: Instead of format, it might be possible to pass:

```
toString={(date, format) => format(new Date(date), 'PP', { locale })}
```

econdepe: Thanks for the suggestion! I did not test much if this new solution is equivalent, but at any rate 'pikaday' will become eventually deprecated by our own library

PR #205, opened 2019-08-20
Fix invalid argument on edge

mmso:
Screenshot 2019-08-20 at 14 14 15

### DIFF
--- a/containers/rightToLeft/Provider.js
+++ b/containers/rightToLeft/Provider.js
@@ -11,7 +11,7 @@ const Provider = ({ children }) => {
         if (isRTL) {
             document.documentElement.dir = 'rtl';
         } else {
-            document.documentElement.dir = '';
+            document.documentElement.dir = 'ltr';
         }
     }, [isRTL]);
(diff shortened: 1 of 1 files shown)


containers/rightToLeft/Provider.js
```diff
@@ -11,7 +11,7 @@ const Provider = ({ children }) => {
         if (isRTL) {
             document.documentElement.dir = 'rtl';
         } else {
-            document.documentElement.dir = undefined;
+            document.documentElement.dir = '';
```
vincaslt: I assumed 'undefined' would not add `dir` at all, but otherwise, I think the only possible values are `rtl` or 'ltr`, not empty (https://www.w3.org/TR/REC-html40/struct/dirlang.html)

Does it not add anything to `dir` attribute, or leaves it empty?

mmso: Turns out it adds an empty attribute 
Screenshot 2019-08-20 at 14 18 41

PR #223, opened 2019-08-26
Hide dropdown on scroll

vincaslt:
If dropdown is supposed to be closed when clicking outside, it's removed on scroll also. Should fix: https://github.com/ProtonMail/protonvpn-settings/issues/69

### DIFF
--- a/components/dropdown/Dropdown.js
+++ b/components/dropdown/Dropdown.js
@@ -1,10 +1,12 @@
-import React, { useEffect, useRef } from 'react';
+import React, { useEffect, useRef, useCallback } from 'react';
 import PropTypes from 'prop-types';
 import keycode from 'keycode';
 import { classnames } from '../../helpers/component';
 import { usePopper, Popper } from '../popper';
 import useRightToLeft from '../../containers/rightToLeft/useRightToLeft';
 import { noop } from 'proton-shared/lib/helpers/function';
+
+const scrollContainerClass = 'main';
 
 const Dropdown = ({
     anchorRef,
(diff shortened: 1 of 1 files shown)


components/dropdown/Dropdown.js
```diff
@@ -61,10 +69,13 @@ const Dropdown = ({
         document.addEventListener('touchstart', handleClickOutside);
         document.addEventListener('keydown', handleKeydown);
 
+        const scrollContainer = document.getElementsByClassName(scrollContainerClass)[0] || document.body;
```
EpokK: I'm not a big fan of using `getElementsByClassName` to select an element in the DOM with React. I prefer if we use a `ref` from the main body or just use `document.body`.

dhoko: also to get the first one querySelector > * ;)

vincaslt: Me too. However, in this case, my thinking is that - we mostly use the same layout components, so the main container is usually `<main className="main main-area>`, and I really don't want to have to pass ref of that container to every dropdown we create. Getting the `main` element by query/className I can set the default to `main` so that we don't need to specify that prop most of the time.
`document.body` is not really our scroll container, and `querySelector` is just syntactic sugar in this case.

Any thoughts?

vincaslt: Another more explicit way I've been thinking about, but which requires a little more effort is to have a wrapper that basically marks the component as scroll container for dropdowns and use it from context as ref.

Maybe something like:

```
// PublicLayout.js
<DropdownScrollContainer>
    {(ref) => <main ref={ref} className="main flex-item-fluid main-area">{children}</main>}
</DropdownScrollContainer>

// Dropdown.js
const { placement, position } = usePopper(popperRef, anchorRef, isOpen, {
    scrollContainer
});

// usePopper.js
[...]
const contentArea = scrollContainer.current || document.body;
[...]
```

mmso: We did something similar to get the sticky title more easily in the settings app, where a ref to the main element is being provided:

https://github.com/ProtonMail/protonmail-settings/blob/master/src/app/components/layout/PrivateLayout.js#L30

https://github.com/ProtonMail/react-components/blob/master/components/container/SettingsTitle.js

It's not super pretty either, but for now it's ok I guess. To use that in the dropdown we would need to enforce all apps to provide a ref to the main element.

vincaslt: Nice, didn't know we had that. It's pretty much what I've described, so I'll use it. This will make it more consistent.

vincaslt: There are a couple of complicated things though with context. It's value would have to be re-defined in several places:

* in most apps it's just main-area, but for instance header, is outside the main area, so it's not part of the context.
* in contacts, we have several scrollable areas - contacts list and contact view.
* modals are outside of any context, so they can't even re-define the ref to be ContentModal (at least I couldn't)
* it affects also tooltips

mmso: Hmm I see.

I think it could make sense to redefine the context for each situation. It should be possible to redefine it for modals also, but I'll investigate that.

Maybe let's wait with this PR a bit.

Issue #229, opened 2019-08-28
All cards are broken visually


EpokK:
![Capture d’écran 2019-08-28 à 15 51 48](https://user-images.githubusercontent.com/1345786/63861771-cf898e80-c9ab-11e9-9e89-86d7bafe6dc4.png)

nico3333fr:
The problem comes from `inline-flex` class applied to `label` that contains cards. A way to make an exception for this case?

Issue #220, opened 2019-08-23
Layout in appearance 


nico3333fr:
The radios with layout choices in Appearance in Settings are not displayed correctly:

![image](https://user-images.githubusercontent.com/2578321/63610179-13118080-c5d9-11e9-9d0d-a3470b7c4e37.png)

while it should be displayed:

![image](https://user-images.githubusercontent.com/2578321/63610480-d09c7380-c5d9-11e9-9a85-3fdd518fdd20.png)

The problem comes from the class `inline-flex` applied by default on `<radio` component. In this special case, it should not be present.

Do I add a modifier to avoid "inline-flexing" in this case or is it easy to update? (by default with `inline-flex` class but possible to remove it using a parameter or something else)

Please advise guys ;)

PR #231, opened 2019-08-29
Fix card display

nico3333fr:
The problem comes from the class `inline-flex` applied by default on `<radio` component. In this special case, it should not be present.

- `inline-flex` class will be added by default is no class is specified

image

Fix https://github.com/ProtonMail/react-components/issues/220 
Fix https://github.com/ProtonMail/react-components/issues/229

### DIFF
--- a/components/input/Radio.js
+++ b/components/input/Radio.js
@@ -5,7 +5,7 @@ import { classnames } from '../../helpers/component';
 
 const Radio = ({ id, children, className = 'inline-flex', ...rest }) => {
     return (
-        <label htmlFor={id} className={classnames([className])}>
+        <label htmlFor={id} className={className}>
             <input id={id} type="radio" className="pm-radio" {...rest} />
             <span className="pm-radio-fakeradio" />
             {children}
--- a/components/input/Radio.js
+++ b/components/input/Radio.js
@@ -1,8 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
-import { classnames } from '../../helpers/component';
-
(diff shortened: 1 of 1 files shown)


components/input/Radio.js
```diff
@@ -3,9 +3,9 @@ import PropTypes from 'prop-types';
 
 import { classnames } from '../../helpers/component';
 
-const Radio = ({ id, children, className, ...rest }) => {
+const Radio = ({ id, children, className = 'inline-flex', ...rest }) => {
     return (
-        <label htmlFor={id} className={classnames(['inline-flex', className])}>
+        <label htmlFor={id} className={classnames([className])}>
```
mmso:
```suggestion
        <label htmlFor={id} className={className}>
```

components/input/Radio.js
```diff
@@ -3,9 +3,9 @@ import PropTypes from 'prop-types';
 
 import { classnames } from '../../helpers/component';
```
EpokK: If you are not using `classnames` in this file, you need to remove this import.

mmso: My bad, was because of my suggested change ^^

EpokK: @mmso 🔫 😄

nico3333fr: I'm fixing it no problem ;)

nico3333fr: Done https://github.com/ProtonMail/react-components/pull/231/commits/a1830410f71b97b164bd1d01a50c382eaae8b3b9

PR #245, opened 2019-09-02
Add logo only on mobile in sidebar container

nico3333fr:
Added logo in nav container. Only for mobile

### DIFF
--- a/components/sidebar/Sidebar.js
+++ b/components/sidebar/Sidebar.js
@@ -2,10 +2,12 @@ import React from 'react';
 import PropTypes from 'prop-types';
 
 import NavMenu from './NavMenu';
+import MainLogo from '../logo/MainLogo';
 
 const Sidebar = ({ list = [] }) => {
     return (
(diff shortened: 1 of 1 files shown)


components/sidebar/Sidebar.js
```diff
@@ -2,10 +2,12 @@ import React from 'react';
 import PropTypes from 'prop-types';
 
 import NavMenu from './NavMenu';
+import MainLogo from '../logo/MainLogo';
 
 const Sidebar = ({ list = [] }) => {
     return (
         <div className="sidebar flex flex-column noprint">
+            <MainLogo url="/account" className="nodesktop notablet" />
```
EpokK: The `url` needs to come from a `prop`. I'm fine with having `/account` String as default.

nico3333fr: Done with https://github.com/ProtonMail/react-components/pull/245/commits/031766ecfdd90dd7be3925312267a89b6c456807

PR #254, opened 2019-09-04
IntlTelInput component

### LINKED ISSUE #250: Phone input component
Create phone input component. The user should be able to select the country flag for the phone extension.

Capture d’écran 2019-09-03 à 15 02 29

An Angular example can be found [here](https://mail.protonmail.com/create/new) on the human verification step.

### Note

- I'm fine to use the same or new library for it.
- Adapt it with @nico3333fr for v4 style.

Yiin:
Fixes #250 

image

I haven't touched the formatting as I'm not sure how it should work. Currently it functions as simple text input with placeholder as an example.

Also the design is not defined as well, so I tried to make it look at least decent for now.

Needs https://github.com/ProtonMail/design-system/pull/219

### DIFF
--- a/components/input/IntlTelInput.js
+++ b/components/input/IntlTelInput.js
@@ -1,0 +1,21 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import ReactIntlTelInput from 'react-intl-tel-input';
+import 'react-intl-tel-input/dist/main.css';
+import { classnames } from '../../helpers/component';
+import './IntlTelInput.scss';
+
+const IntlTelInput = ({ containerClassName, inputClassName, ...rest }) => (
+    <ReactIntlTelInput
+        containerClassName={classnames(['intl-tel-input', containerClassName])}
+        inputClassName={classnames(['pm-field', inputClassName])}
+        {...rest}
+    />
+);
+
+IntlTelInput.propTypes = {
+    containerClassName: PropTypes.string,
+    inputClassName: PropTypes.string
+};
+
+export default IntlTelInput;
--- a/components/input/IntlTelInput.scss
+++ b/components/input/IntlTelInput.scss
@@ -1,0 +1,11 @@
+// Remove selected flag background when dropdown is active
+.intl-tel-input.allow-dropdown .flag-container:hover .selected-flag {
+    background-color: transparent;
+}
+// Apply the background if user hovers the flag directly
+.intl-tel-input.allow-dropdown .flag-container .selected-flag:hover {
+    background-color: rgba(0, 0, 0, 0.05);
+}
+.intl-tel-input .country-list .country.highlight::before {
+    background-color: rgba(0, 0, 0, 0.05);
+}
--- a/package.json
+++ b/package.json
@@ -49,6 +49,7 @@
     "quill": "^1.3.6",
     "react-codemirror2": "^6.0.0",
     "react-color": "^2.17.0",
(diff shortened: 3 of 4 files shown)


mmso:
> styles are imported from the component, not sure if that doesn't cause any problems without css-modules. Maybe at least import 'react-intl-tel-input/dist/main.css'; could be imported from app.scss, but if it only imports once without problems, then nevermind, I'd actually prefer to import from the component.

It only imports it once even if it's imported from the component. And it works as expected, e.g. if this component would be lazily loaded, the css would follow the lazy loading (once). So yeah keep it in the component.

mmso:
I noticed this component is using an inline absolutely positioned select dropdown. This will cause problems for us since we need to portal it.

mmso:
For this https://github.com/ProtonMail/react-components/pull/254#issuecomment-528228700 I think it's better if we'd write our own component. It shouldn't be too complicated.

EpokK:
@Yiin Can you try to build this component by just using the extension list from a lib and use our portal to manage the dropdown?

components/input/IntlTelInput.scss
```diff
@@ -0,0 +1,7 @@
+.intltelinput-input {
+    border-left-color: transparent !important;
```
Yiin: `!important` because dropdown class styles are applied after these styles. A temporary fix to avoid double border between elements.

hooks/useIntlTelNumbers.js
```diff
@@ -0,0 +1,365 @@
+import { useState } from 'react';
+import 'libphonenumber-js-utils';
```
Yiin: This lib is an interesting one.
As said in the README.md of the package (<https://www.npmjs.com/package/libphonenumber-js-utils>), it is a compiled and minified version of the utilities functions from the googlei18n's libphonenumber library. The functions are exposed to window.intlTelInputUtils inside browser environment.

We may want to clone googles lib and use it w/o 3rd parties, but it looks quite complex and I didn't managed to wrap my head around how to use it in reasonable timeframe so I leave it open for a discussion.

Yiin:
Looking at it now, the countries list could be moved to proton-shared maybe?

hooks/useIntlTelNumbers.js
mmso: What do you think about using https://github.com/catamphetamine/libphonenumber-js instead? It seems to have a smaller footprint.

package.json
```diff
@@ -1,94 +1,95 @@
 {
-  "name": "react-components",
```
mmso: Can we revert the style changes on package.json to only have your relevant changes please?

EpokK:
For the dropdown, I would like to change 3 things:

- **order**: `flag` `prefix phone number` `country name`.
- **limit width** of the dropdown and use `ellipsis` in content.
- add a **title** on each item.

components/input/IntlTelInput.js
```diff
@@ -0,0 +1,56 @@
...
+                                <span className="smaller color-global-altgrey">+{dialCode}</span>
+                            </button>
+                        </li>
+                    ))}
+                </ul>
+            </SimpleDropdown>
+            <Input
+                className={classnames(['intltelinput-input rounded0-left', inputClassName])}
+                placeholder={selectedCountry.exampleNumber}
+                {...rest}
```
mmso: Don't we also need to parse the inputted phone number? e.g. if someone enters +44 in the input it should change the flag to british etc.

PR #264, opened 2019-09-09
Fix TOS naming

nico3333fr:
- updated sentence

image

Fixes: https://github.com/ProtonMail/proton-vpn-settings/issues/174


EpokK:
This require some Javascript logic to know if we need to display `Next` or `Upgrade`.

PR #271, opened 2019-09-11
Throw in render if a model or api result fails

mmso:
Fix https://github.com/ProtonMail/proton-vpn-settings/issues/211
Fix https://github.com/ProtonMail/proton-vpn-settings/issues/210

* If a model fails to be fetched, the hook will throw in render
* If a request from `useApiResult` fails to be fetched, the hook will throw in render
* And adds an error boundary to modals
Screenshot 2019-09-11 at 13 35 50

### DIFF
--- a/hooks/useCachedModelResult.js
+++ b/hooks/useCachedModelResult.js
@@ -79,6 +79,11 @@ const useCachedModelResult = (cache, key, miss) => {
     });
     const keyRef = useRef(key);
 
+    if (state && state[2]) {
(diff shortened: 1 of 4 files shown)


hooks/useCachedModelResult.js
```diff
@@ -79,6 +79,11 @@ const useCachedModelResult = (cache, key, miss) => {
     });
     const keyRef = useRef(key);
 
+    if (state && state[2]) {
```
EpokK: Can we make `state[2]` more verbose please?

PR #277, opened 2019-09-11
Fix #276 - loader centering

### LINKED ISSUE #276: Loader image is not centered
Capture d’écran 2019-09-11 à 21 41 05

### Component
``` jsx
<Loader />
```

``` jsx
const Loader = ({ size = 'small' }) => {
    return (
        <div className="center mb2 mt2">
            <img
                src={IMAGES[size]}
                width={size === 'medium' ? MEDIUM_WIDTH : undefined}
                height={size === 'medium' ? MEDIUM_HEIGHT : undefined}
                alt={c('Info').t`Loading`}
            />
        </div>
    );
};
```

nico3333fr:
- centered loader using `flex` and `mauto` 

image

Fixes: #276

### DIFF
--- a/components/loader/Loader.js
+++ b/components/loader/Loader.js
@@ -15,8 +15,9 @@ const MEDIUM_HEIGHT = '80';
 
 const Loader = ({ size = 'small' }) => {
     return (
-        <div className="center mb2 mt2">
+        <div className="center flex mb2 mt2">
             <img
+                className="mauto"
                 src={IMAGES[size]}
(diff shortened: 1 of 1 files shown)


components/loader/Loader.js
```diff
@@ -15,8 +15,9 @@ const MEDIUM_HEIGHT = '80';
 
 const Loader = ({ size = 'small' }) => {
     return (
-        <div className="center mb2 mt2">
+        <div className="center flex mb2 mt2">
```
mmso: just an aligncenter would have been enough instead of using flex no?

PR #284, opened 2019-09-13
Fix unsupported kb url

mmso:
Special case for ProtonVPN

### DIFF
--- a/containers/app/CompatibilityCheck.js
+++ b/containers/app/CompatibilityCheck.js
@@ -93,7 +93,7 @@ const CompatibilityCheck = ({ children }) => {
                     </Href>
                     .
                 </p>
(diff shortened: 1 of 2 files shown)


containers/app/CompatibilityCheck.js
```diff
@@ -85,10 +93,7 @@ const CompatibilityCheck = ({ children }) => {
                     </Href>
                     .
                 </p>
-                <Href
-                    className="primary-link bold"
-                    url="https://protonmail.com/support/knowledge-base/browsers-supported/"
-                >
+                <Href className="primary-link bold" url={kbUrl}>
```
EpokK:
```suggestion
                <Href className="primary-link bold" url={kbUrl} target="_self">
```

PR #285, opened 2019-09-13
Fix - spacing pricing vpn

nico3333fr:
- added `data-suffix` attribute
- added a rule in Design system to add margin when `data-suffix` is not starting by `/` : 

```css
.plan-price-suffix:not([data-suffix^="/"]) {
  margin-left: .5em;
}
```
(no need for space for other cases)

image

image

Fixes: https://github.com/ProtonMail/proton-vpn-settings/issues/241

### DIFF
--- a/components/price/Price.js
+++ b/components/price/Price.js
@@ -12,7 +12,11 @@ const Price = ({ children: amount = 0, currency = '', className = '', divisor = 
     const c = <span className="currency">{CURRENCIES[currency] || currency}</span>;
     const p = value < 0 ? <span className="prefix">-</span> : null;
     const v = <span className="amount">{Math.abs(value)}</span>;
-    const s = suffix ? <span className="suffix ml0-5">{suffix}</span> : null;
+    const s = suffix ? (
+        <span className="plan-price-suffix" data-suffix={suffix}>
+            {suffix}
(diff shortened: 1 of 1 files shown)


EpokK:
@nico3333fr I think we always need spacing for suffix. We should apply something like:

```jsx
{suffix ? <> {s}</> : null}
```

nico3333fr:
@EpokK this was what I did first, but it seemed strange to have a space when we have `/year`

image

If prefer without:

image

EpokK:
I'm fine with that @nico3333fr .

components/price/Price.js
```diff
@@ -12,7 +12,11 @@ const Price = ({ children: amount = 0, currency = '', className = '', divisor =
     const c = <span className="currency">{CURRENCIES[currency] || currency}</span>;
     const p = value < 0 ? <span className="prefix">-</span> : null;
     const v = <span className="amount">{Math.abs(value)}</span>;
-    const s = suffix ? <span className="suffix">{suffix}</span> : null;
+    const s = suffix ? (
+        <span className="plan-price-suffix" data-suffix={suffix}>
```
EpokK:
```suggestion
        <span className="ml0-5">
```

PR #298, opened 2019-09-23
Typescript

vincaslt:
Opened this PR to collect feedback and keep track of cleanup stuff.

Related PRs:

`react-components`(this): https://github.com/ProtonMail/react-components/pull/298   4️⃣
~`proton-pack`: https://github.com/ProtonMail/proton-pack/pull/26/files 1️⃣~
~`proton-i18n`: https://github.com/ProtonMail/proton-i18n/pull/9 2️⃣~ 
~`proton-shared`: https://github.com/ProtonMail/proton-shared/pull/53/files 3️⃣~
`protonmail-settings`: https://github.com/ProtonMail/proton-mail-settings/pull/131 5️⃣ 
Implementing typescript support through babel loaders as opposed to webpack loaders. This will use `'@babel/preset-typescript'` to compile js and ts files.

Caveats:

1. Babel won't do live typecheck. This is not a problem, because editors like VSCode/WebStorm etc. have plugins to support typescript. Just if there are errors in the code, console doesn't inform about them, and still runs the compiled code. This is done to speed up performance. To prevent accidental errors on our main branches, we can should run `tsc` command on commit, which will do the type checking as part of compilation process.

More info: https://iamturns.com/typescript-babel/

2. React typescript files should have `*.tsx` file extension, as it's the standard.

3. Local JS files are recognized and type checked, so when `.js` functions/components are imported in `.ts` files they will throw errors most of the time. The best solution is to rewrite the javascript file to typescript, but sometimes we might not have the time to do that. Typescript can understand types in js files from type comments, so we can provide types or disable checking for those functions, mostly happens with components that have optional props e.g.:

```ts
//JS file
/** @type any */
export const IAmJavascriptCmp = ({ requiredProp, optionalProp }) => {
    return <>{requiredProp}</>;
}

//TS file
<IAmJavascriptCmp prop="test" />
```

`IAmJavascriptCmp` will be marked with an error, as `optionalProp` is assumed to be missing. `@type` override can prevent the errors (but should eventually be removed when files are rewritten in typescript)

4. JS modules need type declarations. Most popular modules have types out of the box or can be installed with `npm install --save -D @types/module-name`. Our projects are setup to recognize typescript files, and do not require declarations for them. However, javascript files are not recognized, so they need declarations. For example, if we have `proton-shared/lib/constants.js` file and import it from `.ts` file: `import { PERMISSIONS } from 'proton-shared/lib/constants';` it will say: `Could not find a declaration file for module 'proton-shared/lib/constants'`.

Best solution is to just rewrite the file to ts quickly, if the file is too large we can use `any` type for more complex declarations and just come back and fix those parts later. If that's not possible for some reason, we can declare a module in `index.d.ts` file:

```
declare module `proton-shared/lib/constants` {
	const value: any
	export = any
}
```

But that should be the last resort.


tsconfig.json
```diff
@@ -0,0 +1,21 @@
+{
+    "compilerOptions": {
```
dhoko: Why is it inside this project?

vincaslt: tsconfig describes a project for typescript, so every project must have it. However, since we want most rules to be consistent, it would be nice to have a base tsconfig and extend it.

https://www.typescriptlang.org/docs/handbook/tsconfig-json.html#configuration-inheritance-with-extends
https://github.com/Microsoft/TypeScript/issues/18865

package.json
```diff
@@ -2,15 +2,14 @@
...
-  "sideEffects": false,
   "scripts": {
     "test": "jest",
     "export": "./makeExport",
     "i18n:validatejs": "proton-i18n validate lint-functions .",
     "i18n:validate": "proton-i18n extract reactComponents && proton-i18n validate",
-    "lint": "eslint containers components --ext .js --quiet --cache",
-    "pretty": "prettier --write $(find lib -type f -name '*.js')"
+    "lint": "eslint containers components --ext .js,.ts,.tsx --quiet --cache",
+    "pretty": "prettier --write $(find lib -type f -name '*.js')",
```
vincaslt: Should probably include .ts(x) files from now on.

package.json
```diff
@@ -2,15 +2,14 @@
   "name": "react-components",
   "version": "1.0.0",
   "description": "",
-  "main": "index.js",
-  "sideEffects": false,
```
dhoko: do not remove this line it's important for the tree-shacking

vincaslt: This was removed by merge conflict, I'll add it back.

dhoko:
> we can should run tsc command on commit

We have lint-staged for this task, you can edit the config here https://github.com/ProtonMail/react-components/blob/master/package.json#L25

.eslintrc.json
```diff
@@ -1,35 +1,43 @@
...
+            "node": {
+                "extensions": [".js", ".ts", ".tsx"]
+            }
+        }
+    },
+    "extends": ["plugin:@typescript-eslint/recommended", "eslint:recommended", "plugin:react/recommended"],
+    "rules": {
+        "@typescript-eslint/explicit-function-return-type": "off",
+        "@typescript-eslint/no-explicit-any": "off",
+        "@typescript-eslint/indent": "off",
```
vincaslt: Indentation is handled by prettier, but it would make sense to have only one place that's responsible for code formatting. I suggest adding prettier integration into eslint, and have eslint do all the checking: https://prettier.io/docs/en/integrating-with-linters.html#eslint

dhoko: as you always commit with prettier it's not really useful imho. And it will slow down eslint, this one being already slow.

Eslint is about rules, prettier about the style. We always run prettier when we commit. It would be redundant and if you have it inside your editor boring imho

vincaslt: Hmm, maybe then it's specific to my workflow only. I like to have auto-format on save, not a big deal for me. The only problem is that some eslint rules overlap with prettier, so we gotta be careful about them, because once prettier formats, eslint might break (that's the reason I disabled the indent rule here). If that becomes a problem - https://github.com/prettier/eslint-config-prettier can ensure we don't have conflicts anyway...

dhoko: > I like to have auto-format on save

Yeah it's editor specific. 

+1 for eslint-config-prettier

package.json
```diff
@@ -82,7 +84,9 @@
     "react-hooks-testing-library": "^0.5.0",
     "react-router-dom": "^4.3.1",
     "react-test-renderer": "^16.8.6",
-    "ttag": "^1.7.0"
+    "react-testing-library": "^7.0.0",
```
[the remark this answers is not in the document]
mmso: Yes `react-testing-library` has been replaced by `@testing-library/react`

PR #300, opened 2019-09-23
Add icon prop to Input

EpokK:
- Wrap input + icon if needed
- Insert icon to the proper place

### DIFF
--- a/components/input/PasswordInput.js
+++ b/components/input/PasswordInput.js
@@ -3,38 +3,20 @@ import { c } from 'ttag';
 import PropTypes from 'prop-types';
 import { Icon } from 'react-components';
 
-import { generateUID, classnames } from '../../helpers/component';
-import useInput from './useInput';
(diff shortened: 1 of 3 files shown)


components/input/PasswordInput.js
```diff
@@ -3,38 +3,20 @@ import { c } from 'ttag';
...
-                    {...rest}
-                    {...handlers}
-                />
+        <Input
+            type={type}
+            disabled={disabled}
+            icon={
                 <button
                     title={type === 'password' ? c('Label').t`Reveal password` : c('Label').t`Hide password`}
                     className="password-revealer inline-flex flex-item-noshrink"
```
mmso: `password-revealer` not needed anymore?

nico3333fr: No, not needed anymore. `right-icon absolute flex` ftw :)

PR #302, opened 2019-09-24
Fix - support drop down caret color on hover

nico3333fr:
Spotted by design: the caret of the support dropdown button did not have the correct color on hover.

- fixed by adding `topnav-icon` class on it

image


RandomAdversary:
@nico3333fr , @EpokK  should this be applied on signup dropdown too? 

Looks like it's the same component, but I don't see the color change.

image

nico3333fr:
@RandomAdversary as far as I know, no. 

This was only pointed out for "dark" standard theme.

Issue #228, opened 2019-08-28
Storage space status text is not visible in light theme


mmso:
<img width="95" alt="Screenshot 2019-08-28 at 14 37 38" src="https://user-images.githubusercontent.com/352607/63856247-6f8dea80-c9a1-11e9-8f94-4dc02b8ab75c.png">

PR #307, opened 2019-09-25
Fix - storage circle chart

nico3333fr:
This component had to be harmonized with Angular/Design system.
Also, the design must come from Design System.

Changes:
- updated SVG (`viewbox`/etc.)
- removed colors/background attributes (managed via classes, easier to fix theming issues like this)
- wrapp icon in a button (in order to trigger the drop down properly by keyboard)
- harmonized code with Angular version

plop

plop2

plop3

Also fixes theming issues:
plop4

Fixes: https://github.com/ProtonMail/react-components/issues/228


containers/sidebar/StorageSpaceStatus.js
```diff
@@ -12,30 +12,33 @@ const StorageSpaceStatus = ({ children }) => {
     const usedPercent = Math.round(UsedSpace / MaxSpace);
     const maxSpaceFormatted = humanSize(MaxSpace);
     const usedSpaceFormatted = humanSize(UsedSpace);
-    const color = usedPercent < 60 ? 'global-success' : usedPercent < 80 ? 'global-attention' : 'global-warning';
+    const color =
+        usedPercent < 60
+            ? 'circle-bar--global-success'
+            : usedPercent < 80
+            ? 'circle-bar--global-attention'
+            : 'circle-bar--global-warning';
```
Yiin: if we're using cicle-bar prefix, maybe it would be better to name status to low/medium/full instead of color names? Feels a bit redundant otherwise.

nico3333fr: Yes, good idea. :)

As I'm waiting some design elements (icon for the "i"), I'll rename them at the same time when I will update it :)

I've created an issue to avoid forgetting it https://github.com/ProtonMail/design-system/issues/228

PR #313, opened 2019-09-30
Add MainContainer component

Yiin:
Required for https://github.com/ProtonMail/proton-vpn-settings/issues/246

### DIFF
--- a/components/container/MainContainer.js
+++ b/components/container/MainContainer.js
@@ -1,0 +1,27 @@
+import React, { useEffect, useRef } from 'react';
+import PropTypes from 'prop-types';
+import { withRouter } from 'react-router-dom';
+
+const MainContainer = ({ children, location, rootRef }) => {
+    const fallbackRef = useRef();
+    const containerRef = rootRef || fallbackRef;
+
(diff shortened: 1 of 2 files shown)


components/container/MainContainer.js
```diff
@@ -0,0 +1,27 @@
+import React, { useEffect, useRef } from 'react';
+import PropTypes from 'prop-types';
+import { withRouter } from 'react-router-dom';
+
+const MainContainer = ({ children, location, rootRef }) => {
+    const fallbackRef = useRef();
+    const containerRef = rootRef || fallbackRef;
```
econdepe: I wonder if the `useMainArea` hook would be useful for managing this `ref`

components/container/MainContainer.js
```diff
@@ -0,0 +1,27 @@
+import React, { useEffect, useRef } from 'react';
+import PropTypes from 'prop-types';
+import { withRouter } from 'react-router-dom';
+
+const MainContainer = ({ children, location, rootRef }) => {
```
mmso: Why is it called `rootRef`?

Issue #303, opened 2019-09-24
Focus via keyboard doesn't work with the dropdwon


dhoko:
1. Open ex settings/subscription
2. Go to invoices
3. Open via the keyboard the pagination, it's impossible to change the page via the keyboard (_select a number_)

![output](https://user-images.githubusercontent.com/713283/65521376-6c771300-dee9-11e9-9ef8-4fd71a813212.gif)

PR #314, opened 2019-09-30
Focus dropdown

Yiin:
Fixes https://github.com/ProtonMail/react-components/issues/303

### DIFF
--- a/components/popper/Popper.js
+++ b/components/popper/Popper.js
@@ -1,11 +1,29 @@
-import React from 'react';
+import React, { useEffect } from 'react';
 import PropTypes from 'prop-types';
 import Portal from '../portal/Portal';
 
 const Popper = React.forwardRef(({ children, position, isOpen, role = 'tooltip', ...rest }, ref) => {
+    useEffect(() => {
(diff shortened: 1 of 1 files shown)


mmso:
Where does focus end up after closing the dropdown? Is the user able to tab to the next focusable element?

nico3333fr:
If we focus using Maj+tab, do the focus comes back to previous element ?

mmso:
I am starting to think it was perhaps a mistake to move dropdowns into absolutely positioned elements. It did solve the overflow problem, but I feel like it may have caused more problems than it solved. Perhaps we could have solved the overflow problem in another way.

vincaslt:
I agree... It seems to solve the problem in basic cases, but is pretty hard to ensure it works with all the custom stuff we have - modals, non-body scroll containers (especially multiple).

I would like to have this solved in css if possible (from my knowledge it's not doable, unless something new was added that made it possible) or use some 3rd party that handles this for us and tries to catch the edge cases, like https://popper.js.org/ (not sure if it wouldn't run into the same problems that we have now)

PR #317, opened 2019-09-30
Remove themes

econdepe:
Remove light and blue themes from the possible theme choices. These themes are not ready for integration with the Angular mail app.

Also, change a bit the way the themes were handled:
* Simplify a bit the logic, which was a bit scattered before. Now there is one file defining all the themes and helpers in `proton-shared` (see https://github.com/ProtonMail/proton-shared/pull/54, which must be merged first). Here we just need to grab the themes we want to display.
* Localization was not working for the theme labels in the previous version.
* Custom theme modal was not closing after saving the new CSS.
* Button for saving custom CSS is disabled now for empty input. This was generating some slight misbehavior before when saving empty CSS.
* Class missing in `RadioCard` spotted by @nico3333fr.
* Also by @nico3333fr , remove commented code that was used as `id` in the theme cards.
* Remove to-be-deprecated `useApiWithoutResult`.

Needs https://github.com/ProtonMail/proton-shared/pull/54


containers/themes/CustomThemeModal.js
```diff
@@ -1,23 +1,25 @@
...
+import { FormModal, PrimaryButton, Label, Alert, TextArea } from 'react-components';
 
-const CustomThemeModal = ({ onClose, onSave, theme: initialTheme = '', ...rest }) => {
+const CustomThemeModal = ({ onSave, theme: initialTheme = '', ...rest }) => {
     const [theme, setTheme] = useState(initialTheme);
 
     const handleChange = ({ target }) => setTheme(target.value);
-    const handleSubmit = () => onSave(theme);
+    const handleSubmit = async () => {
+        await onSave(theme);
```
EpokK: Can we add a success notification here?

containers/themes/ThemesSection.js
```diff
@@ -3,45 +3,49 @@ import { c } from 'ttag';
...
+    });
+
+    const handleChangeTheme = async (themeIdentifier) => {
+        if (themeIdentifier === CUSTOM.identifier) {
             return handleOpenModal();
         }
-        await request(themeId);
-        call();
+        await api(updateTheme(themeIdentifier));
+        await call();
```
EpokK: Again, can we add a success notification here?

containers/themes/ThemesSection.js
```diff
@@ -3,45 +3,49 @@ import { c } from 'ttag';
...
+import { getThemeIdentifier, stripThemeIdentifier } from 'proton-shared/lib/themes/helpers';
+import { ALL_THEMES } from 'proton-shared/lib/themes/themes.js';
 
-import { availableThemes } from './availableThemes.js';
 import CustomThemeModal from './CustomThemeModal.js';
 
-const {
-    CUSTOM: { identifier: customId }
-} = THEMES;
+const { DARK, CUSTOM } = ALL_THEMES;
```
EpokK: Since we plan to add a "dark" theme later, can we rename `DARK` to `NORMAL`?

containers/themes/CustomThemeModal.js
econdepe: The notification for the custom theme is defined in `ThemesSection`. Like that we avoid calling `useNotifications` here

containers/themes/ThemesSection.js
econdepe: Yep, added!

containers/themes/ThemesSection.js
econdepe: Changed to default

PR #321, opened 2019-10-02
Cache card details

Yiin:
Fixes https://github.com/ProtonMail/proton-vpn-settings/issues/290


containers/payments/Method.js
```diff
@@ -13,9 +13,11 @@ import Bitcoin from './Bitcoin';
 const { CARD, PAYPAL, BITCOIN, CASH } = PAYMENT_METHOD_TYPES;
 
 const Method = ({ type, amount = 0, currency, onCard, onPayPal, method, methods, loading }) => {
-    const { card, updateCard, errors, isValid } = useCard();
+    const cache = useCache();
+    const { card, updateCard, errors, isValid } = useCard(cache.get('card'));
 
     useEffect(() => {
+        cache.set('card', card);
```
mmso: Could we find another solution than using `cache` for this?

vincaslt:
Why is is it needed to cache card details on our end? Browsers do that by default if they detect a card input field: 
image

containers/payments/Method.js
Yiin: I guess we could cache card data in a variable outside the component, but I'm not sure if this PR is even relevant to the project for the reasons @vincaslt mentioned.

containers/payments/Method.js
```diff
@@ -12,8 +12,9 @@ import Bitcoin from './Bitcoin';
 
 const { CARD, PAYPAL, BITCOIN, CASH } = PAYMENT_METHOD_TYPES;
 
-const Method = ({ type, amount = 0, currency, onCard, onPayPal, method, methods, loading }) => {
-    const { card, updateCard, errors, isValid } = useCard();
+const Method = ({ type, amount = 0, currency, onCard, onPayPal, method, methods, loading, cardToUse }) => {
+    const fallbackCard = useCard();
```
mmso: Why is `fallbackCard` needed now? Can't we force `card` to always be passed?

emnproton:
Verified as fixed.

PR #324, opened 2019-10-02
Rename button  when adding credit card verification fails

Yiin:
Fixes https://gitlab.protontech.ch/ProtonVPN/web/protonvpn-settings/issues/7

I'm not sure if passing a prop is the best approach here, but couldn't think of anything better that wouldn't require bigger refactoring.


EpokK:
Has been fixed in d5795d3

PR #330, opened 2019-10-04
Components for the top navigation bar

econdepe:
In order to have consistency across `contacts`, `settings`, and eventually `mail`, I created some components here that can be shared across later.

Some CSS classes have been also fixed thanks to @nico3333fr.


components/link/TopNavbarLink.js
```diff
@@ -0,0 +1,28 @@
...
+        <Link to={to} {...rest}>
+            {iconComponent}
+            <span className="navigation-title topnav-linkText">{text}</span>
+        </Link>
+    );
+};
+
+TopNavbarLink.propTypes = {
+    to: PropTypes.string.isRequired,
+    icon: PropTypes.string.isRequired,
```
EpokK: Why not using `node` here? I don't understand why we need to handle node and string with `iconComponent`.

containers/search/Searchbox.js
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
+import { c } from 'ttag';
+import PropTypes from 'prop-types';
+import { SearchInput, Icon } from 'react-components';
+
+import { classnames } from '../../helpers/component';
+
+const Searchbox = ({ className = '', placeholder = '', value = '', onSearch }) => {
+    return (
+        <div className={classnames(['searchbox-container relative flex-item-centered-vert', className])}>
```
EpokK: Where do you define the "loupe" icon?

components/link/TopNavbarLink.js
econdepe: I just wanted to imitate what we have in the `Button` component, where `icon` can be either a string (name of the icon) or a node (icon itself). But I forgot to modify the PropType accordingly. I did that with the last push.

containers/search/Searchbox.js
```diff
@@ -0,0 +1,36 @@
...
+                    onChange={onSearch}
+                    id="global_search"
+                    placeholder={placeholder}
+                    className="searchbox-field"
+                />
+            </label>
+            <button type="button" className="searchbox-search-button flex">
+                <Icon name="search" className="fill-white mauto searchbox-search-button-icon" />
+                <span className="sr-only">{c('Action').t`Search`}</span>
+            </button>
```
EpokK: You should add also support for an `advanced` button.
Capture d’écran 2019-10-07 à 10 49 25

containers/search/Searchbox.js
EpokK: Can we use a `form` here? And then use `onSubmit={onSearch}` and `type=submit` on `searchbox-search-button`.

components/link/TopNavbarLink.js
EpokK: I don't see the point of using both, just use `PropTypes.string` please.

PR #333, opened 2019-10-06
Add proptypes to confirm modal

econdepe:
`confirm` and `cancel` should have the same propTypes as `submit` and `cancel` in `FormModal`

### DIFF
--- a/components/modal/Confirm.js
+++ b/components/modal/Confirm.js
@@ -35,8 +35,8 @@ Confirm.propTypes = {
     onConfirm: PropTypes.func,
     title: PropTypes.string.isRequired,
     children: PropTypes.node.isRequired,
-    cancel: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),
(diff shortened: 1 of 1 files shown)


components/modal/Confirm.js
```diff
@@ -35,8 +35,8 @@ Confirm.propTypes = {
     onConfirm: PropTypes.func,
     title: PropTypes.string.isRequired,
     children: PropTypes.node.isRequired,
-    cancel: PropTypes.string,
-    confirm: PropTypes.string,
+    cancel: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),
```
EpokK: If you use `PropTypes.node`, it will support also `String` as input.
```suggestion
    cancel: PropTypes.node,
```

econdepe: Yeah, that's true. I thought it was more clear to be that explicit (we do that in some other places). But there's no point in being redundant. Just having `node` now

PR #336, opened 2019-10-08
Rewrite button, pagination, classnames to ts

vincaslt:
Need these for admin panel.


components/button/Button.tsx
```diff
@@ -1,21 +1,25 @@
-import React from 'react';
-import PropTypes from 'prop-types';
+import * as React from 'react';
```
vincaslt: Will change the imports to default, because we use `useSyntheticDefaultImports` flag

PR #361, opened 2019-10-17
Fix - signature rich edit block width

nico3333fr:
- added a modifier to make the Rich Text Edition full width
- also fixed alignement issue

image


containers/addresses/EditAddressModal.js
```diff
@@ -62,9 +62,9 @@ const EditAddressModal = ({ onClose, address, ...rest }) => {
             </Row>
             <Row>
                 <Label>{c('Label').t`Signature`}</Label>
-                <Field>
+                <div className="pm-field-container--full">
```
EpokK:
```suggestion
                <Field className="pm-field-container--full">
```

nico3333fr:
@EpokK just checked, no problem using `Field` :)

PR #376, opened 2019-10-23
Modify useActiveBreakpoints hook

econdepe:
It's convenient to have it return Booleans `isMobile`, `isTablet` and `isDesktop` apart from the breakpoint itself


hooks/useActiveBreakpoint.js
```diff
@@ -23,7 +23,15 @@ const useActiveBreakpoint = () => {
         };
     }, []);
 
-    return breakpoint;
+    return useMemo(
+        () => ({
+            breakpoint,
+            isDesktop: breakpoint === 'desktop',
```
EpokK: Use constants defined in `proton-shared`.

PR #380, opened 2019-10-24
Fix warning in DropdownButton

econdepe:
Sometimes a warning was being thrown in the console because the prop `loading` was passed to `<button>` throught the `DropdownButton` component.


emnproton:
Verified as fixed.

PR #381, opened 2019-10-24
Add confirm modal to switch team from custom

EpokK:
Fix https://github.com/ProtonMail/proton-mail-settings/issues/186

### DIFF
--- a/containers/themes/ThemesSection.js
+++ b/containers/themes/ThemesSection.js
@@ -9,7 +9,8 @@ import {
     Alert,
     ThemeCards,
     useModals,
-    useNotifications
+    useNotifications,
+    ConfirmModal
 } from 'react-components';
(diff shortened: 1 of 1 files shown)


containers/themes/ThemesSection.js
```diff
@@ -44,11 +45,26 @@ const ThemesSection = () => {
...
+            await new Promise((resolve, reject) => {
+                createModal(
+                    <ConfirmModal
+                        submit={c('Action').t`Apply`}
+                        title={c('Title').t`Change theme`}
+                        onConfirm={resolve}
+                        onClose={reject}
+                    >
+                        <Alert type="warning">{c('Warning')
+                            .t`This action will permanently override your current theme. Are you sure you want to apply this new theme?`}</Alert>
```
econdepe: Maybe it would be better `This action will erase your current theme` ? So that it's even more clear that your theme would be lost?

PR #384, opened 2019-10-25
Fix #9219 - address cut off

nico3333fr:

Address cut of due to line-height

- added some classes in Design system and revamped line-height management (harmonized with React/Design system, was the mess)

image


emnproton:
Verified as fixed.

PR #386, opened 2019-10-25
Change wildcard toggle condition and text

econdepe:
The new design require to have the text "Require exact match" for the toggle in `Settings > General > Search` instead of the v3 "Do not require exact match". We simply need to use the opposite values in this toggle. The corresponding PR fix in `mail` is awaiting review

Closes https://github.com/ProtonMail/proton-mail-settings/issues/191


emnproton:
Verified as fixed.

PR #389, opened 2019-10-28
Force image display in RemoteImage component

econdepe:
In some places we may need to force display (like for this issue here: https://github.com/ProtonMail/proton-contacts/issues/231)


Yiin:
We can use  tag. RemoteImage component loses it's point in cases where we want to force display the image.

PR #391, opened 2019-10-28
Change subscriptions option message ids to fix the variable

Swiip:
Fix https://github.com/ProtonMail/proton-mail-settings/issues/193

Translations fail because the singular version doesn't include the variable. It works in the native language because it's a pure JS variable interpolation but for other language, the ttag has to do the interpolation but doesn't know about the variable.


Swiip:
Addition: I didn't find any other ``msgid`1`` in any project so it should be the only occurence of this problem

PR #392, opened 2019-10-28
Handle failure on add filter modal

Swiip:
Fix https://github.com/ProtonMail/proton-mail-settings/issues/197

Handle request failure in filter creation. Close the modal and update the list.

### DIFF
--- a/containers/filters/AddFilterModal.js
+++ b/containers/filters/AddFilterModal.js
@@ -39,12 +39,17 @@ function AddFilterModal({ filter, type, mode = 'create', onEdit = noop, ...props
     const reqUpdate = useApiWithoutResult(updateFilter);
 
     const create = async (filter) => {
-        const { Filter } = await reqCreate.request(filter);
-        call();
-        createNotification({
-            text: c('Notification').t`${Filter.Name} created`
-        });
-        props.onClose();
+        try {
+            const { Filter } = await reqCreate.request(filter);
+            createNotification({
+                text: c('Notification').t`${Filter.Name} created`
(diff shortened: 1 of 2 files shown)


containers/filters/AddFilterModal.js
```diff
@@ -38,12 +38,15 @@ function AddFilterModal({ filter, type, mode = 'create', onEdit = noop, ...props
...
-            text: c('Notification').t`${Filter.Name} created`
-        });
-        props.onClose();
+        try {
+            const { Filter } = await reqCreate.request(filter);
+            createNotification({
+                text: c('Notification').t`${Filter.Name} created`
+            });
+        } finally {
+            call();
```
EpokK: If the request fail we should not call the event manager since there is no filter model to synchronize.

Swiip: It's what I thought at the beginning but in fact yes.

In fact, the filter is added but not activated and you're not allow to activate it. But it's added so without that, it's not shown.

EpokK: It sounds like a bug, you should check the issue with @bartbutler.

Swiip: I just checked with @bartbutler.

It works as expected. The limit is on the number of active filters not "existing" filters. This way, someone spending time to configure a new filter doesn't loose his work.

EpokK: @Swiip could you add a comment in the code explaining why we are calling `call` from the event manager even if the request fail please?

PR #393, opened 2019-10-28
Allow to change the text in the RemoteImage component

econdepe:
I need to use different text than `Load image` in `contacts` (https://github.com/ProtonMail/proton-contacts/issues/231) when using the `RemoteImage` component.


emnproton:
Verified as fixed.

PR #395, opened 2019-10-28
Show fingerprints in title and add a copy button

Swiip:
Fix https://github.com/ProtonMail/proton-mail-settings/issues/202

Tooltip is not good in this case because it doesn't allow to select the fingerprint. In addition, the tooltip default size is not large enough to show the fingerprint.

So I propose a native HTML title and a copy button.

Capture d’écran 2019-10-28 à 17 54 28

### DIFF
--- a/containers/keys/KeysTable.js
+++ b/containers/keys/KeysTable.js
@@ -33,7 +33,7 @@ const KeysTable = ({ keys = [], onAction }) => {
                             key={ID}
                             cells={[
                                 <div key={1} className="flex flex-row flex-nowrap flex-items-center">
-                                    <Copy value={fingerprint} />
+                                    <Copy value={fingerprint} className="pm-button--small" />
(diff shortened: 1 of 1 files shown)


containers/keys/KeysTable.js
```diff
@@ -31,9 +32,12 @@ const KeysTable = ({ keys = [], onAction }) => {
                         <TableRow
                             key={ID}
                             cells={[
-                                <code key={1} className="mw100 inbl ellipsis">
-                                    {fingerprint}
-                                </code>,
+                                <div key={1} className="flex flex-row flex-nowrap flex-items-center">
+                                    <Copy value={fingerprint} />
```
EpokK: Make `<Copy>` smaller:
```suggestion
                                    <Copy value={fingerprint} className="pm-button--small" />
```

Swiip: Thought that was a bit big too, doesn't know about this class. Cool.

PR #400, opened 2019-10-29
Fix SIEVE editor by delaying content

Swiip:
Fix https://github.com/ProtonMail/proton-mail-settings/issues/214

This is a tricky bug inside codemirror. The selection is done by measuring and simulating the code area.

Perhaps due to the modal appearance animation, the initial measures are corrupted. Any refresh after that fix the problem. Any edition of the code does too.

The problem was: when to trigger a render? I did not found any good events to connect to. I never recommand to use setTimeout but I did not found anything else here.

"Theoretically" the measure issue is due to CSS annimation so the time is not directly dependant on the user computer performance so it "should" be stable.

Edit: the exact timing is 200ms, I used 300 to be secure.

### DIFF
--- a/containers/filters/editor/Sieve.js
+++ b/containers/filters/editor/Sieve.js
@@ -1,4 +1,4 @@
-import React, { useEffect } from 'react';
+import React, { useEffect, useState } from 'react';
 import codemirror from 'codemirror';
 import PropTypes from 'prop-types';
 import { UnControlled as CodeMirror } from 'react-codemirror2';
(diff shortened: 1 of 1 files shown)


containers/filters/editor/Sieve.js
```diff
@@ -63,7 +67,7 @@ function FilterEditorSieve({ filter, onChangeBeforeLint = noop, onChange = noop
     return (
         <CodeMirror
             className="bordered-container"
-            value={filter.Sieve}
+            value={value}
```
EpokK: And the `onChange` prop doesn't need to interact with the `setValue` function?

Swiip: Good remark which made me read a bit about [uncontrolled](https://reactjs.org/docs/uncontrolled-components.html) and [controlled](https://reactjs.org/docs/forms.html#controlled-components) components.

As we use the uncontrol version of codemirror, this seems to work well because the value is only used at initialization. In controlled version, it would indeed fail.

But alternatively, I could also not messing with the value. Considering my tests, any React render will fix the problem. I can create a dummy state to artificially trigger a render. But as it looked strange, I prefered link this to an actual value.

PR #403, opened 2019-10-30
Handle empty signature to send real empty string

Swiip:
Fix https://github.com/ProtonMail/proton-mail-settings/issues/177

Intercept empty input from the reach text editor for the signatuer and replace it by a real empty string.

### DIFF
--- a/containers/addresses/EditAddressModal.js
+++ b/containers/addresses/EditAddressModal.js
@@ -15,9 +15,9 @@ import {
     useEventManager
 } from 'react-components';
 
-const RichTextEditorEmptyValue = '<div><br></div>';
+const EMPTY_VALUES = [/<div><br><\/div>/, /<div>\s*<\/div>/];
 
-const formatSignature = (value) => (value === RichTextEditorEmptyValue ? '' : value);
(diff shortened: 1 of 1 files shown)


containers/addresses/EditAddressModal.js
```diff
@@ -15,6 +15,10 @@ import {
     useEventManager
 } from 'react-components';
 
+const RichTextEditorEmptyValue = '<div><br></div>';
```
EpokK: What do you think about putting it in an Array constant?
```suggestion
const EMPTY_VALUES = ['<div><br></div>'];
```

containers/addresses/EditAddressModal.js
```diff
@@ -15,6 +15,10 @@ import {
...
+
+const formatSignature = (value) => (value === RichTextEditorEmptyValue ? '' : value);
```
EpokK: And then:
```suggestion
const formatSignature = (value) => (EMPTY_VALUES.includes(value)  ? '' : value);
```

emnproton:
Verified as fixed.

PR #404, opened 2019-10-30
Fix missing error message in card expiration field

Swiip:
Fix https://github.com/ProtonMail/proton-mail-settings/issues/161

Validation seems to work. Was just missing a required attribute to handle when the user never touched the field.


emnproton:
Verified as fixed.

PR #405, opened 2019-10-30
Fix settings dashboard anchor links

Swiip:
(Almost) Fix https://github.com/ProtonMail/proton-mail-settings/issues/159

The links in the dashboard was containing the hash but in a soft navigation like this, it doesn't work.

I chose to change the link to add a manual `scrollIntoView` after the navigation.

It works pretty well unless that the scroll command is ran just after the navigation and if the page has async content, it will scroll to the position the anchor was before the data are there.

But to be honest, I don't see how it is possible to wait until the page is ready. So at least, it's better as before.

### DIFF
--- a/containers/overview/IndexSection.js
+++ b/containers/overview/IndexSection.js
@@ -5,7 +5,7 @@ import { classnames } from '../../helpers/component';
 
 import Sections from './Sections';
 
-const IndexSection = ({ pages, history }) => {
+const IndexSection = ({ pages }) => {
     const permissions = usePermissions();
     return (
         <div className="settings-grid-container">
@@ -18,7 +18,7 @@ const IndexSection = ({ pages, history }) => {
                         <h2 className="h6 mb0-5">
                             <strong>{text}</strong>
                         </h2>
-                        {Sections({ route, sections, text, permissions, pagePermissions, history })}
+                        {Sections({ route, sections, text, permissions, pagePermissions })}
                     </div>
                 );
             })}
@@ -27,8 +27,7 @@ const IndexSection = ({ pages, history }) => {
 };
 
 IndexSection.propTypes = {
-    pages: PropTypes.array,
-    history: PropTypes.object.isRequired
+    pages: PropTypes.array
 };
 
 export default IndexSection;
--- a/containers/overview/IndexSection.js
+++ b/containers/overview/IndexSection.js
@@ -5,7 +5,7 @@ import { classnames } from '../../helpers/component';
 
 import Sections from './Sections';
 
-const IndexSection = ({ pages }) => {
+const IndexSection = ({ pages, history }) => {
     const permissions = usePermissions();
     return (
         <div className="settings-grid-container">
@@ -18,7 +18,7 @@ const IndexSection = ({ pages }) => {
                         <h2 className="h6 mb0-5">
                             <strong>{text}</strong>
                         </h2>
-                        {Sections({ route, sections, text, permissions, pagePermissions })}
+                        {Sections({ route, sections, text, permissions, pagePermissions, history })}
                     </div>
                 );
             })}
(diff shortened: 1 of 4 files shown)


containers/overview/LinkItem.js
```diff
@@ -1,26 +1,41 @@
...
-    route: PropTypes.string,
+    route: PropTypes.object,
     permission: PropTypes.bool,
-    text: PropTypes.string
+    text: PropTypes.string,
+    history: PropTypes.object
 };
 
-export default LinkItem;
+export default withRouter(LinkItem);
```
EpokK: Instead of wrapping `LinkItem` with `withRouter`, what do you think about passing `history` by prop from his parent component: `Sections`?

Swiip: It's what I did in the first place. It requires to send it down through 4 components and need a PR on the https://github.com/ProtonMail/proton-mail-settings project so I thought it was a bit painful. But totally possible, as you want.

containers/overview/LinkItem.js
```diff
@@ -1,26 +1,41 @@
...
+
+        const el = document.querySelector(`#${route.hash}`);
+        if (el) {
+            el.scrollIntoView({ behavior: 'smooth', block: 'start' });
+        }
+    };
+
     return (
-        <Link to={route}>
+        <a onClick={handleClick} href={`${route.pathname}#${route.hash}`}>
```
econdepe: Maybe we can use the `Href` component here?

containers/overview/LinkItem.js
econdepe: That kind of swingling `history` around is done in many places in `contacts`, so feat not! :sweat_smile:

containers/overview/LinkItem.js
```diff
@@ -1,26 +1,41 @@
...
+import { withRouter } from 'react-router-dom';
 import { Tooltip, Icon } from 'react-components';
 import { c } from 'ttag';
 
-const LinkItem = ({ route, text, permission }) => {
+const LinkItem = ({ route, text, permission, history }) => {
+    const handleClick = async (event) => {
+        event.preventDefault();
+
+        history.push(route.pathname);
```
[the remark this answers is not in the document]
Swiip: Yes

Swiip:
Done but need a PR on https://github.com/protonmail/proton-mail-settings

containers/overview/IndexSection.js
```diff
@@ -27,7 +27,8 @@ const IndexSection = ({ pages }) => {
 };
 
 IndexSection.propTypes = {
-    pages: PropTypes.array
+    pages: PropTypes.array,
+    history: PropTypes.object
```
econdepe: I would add `history: PropTypes.object.isRequired` here and in the other places used

Swiip: You're right.

containers/overview/LinkItem.js
```diff
@@ -1,26 +1,42 @@
...
+
+        const el = document.querySelector(`#${route.hash}`);
+        if (el) {
+            el.scrollIntoView({ behavior: 'smooth', block: 'start' });
+        }
+    };
+
     return (
-        <Link to={route}>
+        <a onClick={handleClick} href={href(route)}>
```
mmso: I have another proposal than this.

1. Revert this to be normal links, with a hash anchor.
2. Do the `scrollIntoView` in the `Page.js` file in a useEffect by observing the hash instead.

It still wouldn't solve the dynamic section issue, but at least it has the advantage of working also on page refresh, and it doesn't break right-click > open in a new tab (even though we don't support that atm due to session not being shared).

Swiip: You're right, it would be much easier to maintain. I'll try this.

Swiip: Working on this, I seen that the observer sidebar also do a scroll on click instead of changing the hash and react to it. Perhaps, I could change this too ?

Swiip:
Ok! I almost rollback everything to let the scroll be done by the `Page` component itself.

Once that done, I change the sidebar behavior to use the same system: changing the hash in the URL and let the Page do the scroll.

Linked to https://github.com/ProtonMail/proton-mail-settings/pull/225

PR #407, opened 2019-10-31
Remove search input init in each search to keep trace of the active s…

Swiip:
Fix https://github.com/ProtonMail/proton-contacts/issues/275

Simply remove the state reset to keep trace of the previous search.


Swiip:
Add updating the search value on the prop value to be up to date when search has been changed from outside.

PR #413, opened 2019-11-01
Fix - mini calendar display

nico3333fr:

Display/revamping of mini-calendar.

- removed "pure" inline-styles to move them to Design System, used CSS custom properties instead
- removed other inline-styles, replaced by a modifier `mini-calendar-grid--displayWeekNumber`
- removed gridSize (transferred to Design System, always equal to `1fr`)
- fixed a bug on `div class="minicalendar"` (was `aria-label="[object object]"` instead of text, see fix applied image
- fix icon display
- added some stuff for a11y

(as styles did not have impact on other apps, they are already committed on Design system)

## Snapshots

Display without week number: (same as Zeplin)
image

Display with week number (there was not yet any design for it, so tried something to... start with something, will enhance later with designers)

image


nico3333fr:
On hold, the input date has some bugs to fix before merge.

nico3333fr:
Well, this should manage the background issue now.

PR #415, opened 2019-11-05
Fix false positives on signature ending with empty lines

Swiip:
Fix https://github.com/ProtonMail/proton-mail-settings/issues/228

Sorry, that was an error in my regex. Should be stronger now.


emnproton:
Verified as fixed.

PR #416, opened 2019-11-05
Remove debounce on sieve lint

Swiip:
Fix https://github.com/ProtonMail/proton-mail-settings/issues/221

The PR is very simple yet I dug a bit. The problem is that we added a debounce on the lint event to save some hit to the API. But this debounce lose the return value of the lint function which codemirror use to display the eventual error.

I worked on a "promised based version of the debounce" to make it works preserving the result. But while debugging it, I discover that codemirror already has a debounce mechanism and even if there is a lot of calls, the additional debounce doesn't catch much calls (At most, one or two calls in a whole normal usage).

So clearly, this simplest solution is to remove it.

(In Angular, there is no additional debounce either)


emnproton:
Verified as fixed.

PR #424, opened 2019-11-07
Fix #423 - minicalendar layout

### LINKED ISSUE #423: Date picker bugs & quick wins
- [x] Month/year should be centered
- [x] Arrows to move to previous month/next month should be visible (i cannot select another month unless I manually replace `Nov` by `Dec` or `Jan` in the text input
- [x] today's date could be highlighted with primary blue
- [x] clicking on a date does not select it and close the date picker (it would probably require a click effect and then automatically close)
- [x] add padding

here is the design direction: https://app.zeplin.io/project/5d3afcab62259b6cb60d7059/screen/5dbc40ba03012e5571e751ff

here is what I currently experience in FF on Mac OSX
Capture d’écran 2019-11-06 à 20 36 48

nico3333fr:

- reordered/re-aligned buttons
- resize cells
- fix rounded issue for current day
- styled markers properly

Fixes #423 

## Snapshot

image

image

### DIFF
--- a/components/miniCalendar/MiniCalendar.js
+++ b/components/miniCalendar/MiniCalendar.js
@@ -73,22 +73,36 @@ const MiniCalendar = ({
             onMouseDown={preventLeaveFocus}
             aria-label={monthLabel}
         >
-            <div className="flex">
-                <span className="bold flex-item-fluid">{monthLabel}</span>
+            <div className="flex flex-items-center p1">
                 {hasCursors ? (
                     <>
-                        <button type="button" title={prevMonth} className="mr1" onClick={() => handleSwitchMonth(-1)}>
+                        <button
+                            type="button"
+                            title={prevMonth}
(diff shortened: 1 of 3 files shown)


mmso:
@nico3333fr I noticed a strange behavior if you click on the inner element to drag-to-select and select a date. I think one solution is to add pointer events none on the new inner child at `minicalendar-day-inner`. And did you test to display it without week numbers?

nico3333fr:
@mmso good point, I added `no-pointer-events-children` on button, to be sure nothing inside will react.

It works well without week number (well, in fact, the difficulties were more with week numbers, without there is much free room).

image

mmso:
Yeah seems good now. It seems like the rounded circle becomes a bit more elliptic without week numbers?

nico3333fr:
Maybe pull design system, it should be round :)

PR #427, opened 2019-11-12
Improvements on filter modal

Swiip:
Fix: https://github.com/ProtonMail/proton-mail-settings/issues/236

For the first point about deleting a condition. What I seen is that the model data in React was ok but some `select` was not displaying what they should because of the tracking of conditions based on the array index. So I added a computed id on conditions and it should be better.

For the last point about adding value, waiting for feedback.

### DIFF
--- a/containers/filters/FilterConditionValues.js
+++ b/containers/filters/FilterConditionValues.js
@@ -16,6 +16,24 @@ function FilterConditionValues({
     onEdit = noop,
     error = {}
 }) {
+    const allButLast = condition.Values.slice(0, -1);
+    const lastIndex = condition.Values.length - 1;
+    const last = condition.Values[lastIndex];
+
+    const handleEdit = (index) => (value) => {
+        if (index === -1) {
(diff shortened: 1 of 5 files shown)


Swiip:
For the last point, after checking with @matstaz, I changed the behavior to consider the last line as an active value. The Add button now add a line but the value is already in the model. The only drawback is that if you use the form as before, you can end up with an empty value without knowing it.

Working with values, I discovered a bug. Edition and deletion was indexed on the actual value. It meant that having multiple times the same values resulted in editing or deleting all the values at once. I fixed that by using array indexes instead.

Swiip:
I have an unpushed commit triggering the validation on all inputs after the first submit. This way, errors are disapearing while the user correct the form.

The limitation is that it does the validation at every keystroke, not on blur. That would require a lot more code.

Waiting for validation to push (or not) this commit.

Swiip:
Seen with @matstaz. We can merge these changes and talk about form validation harmonization later.

Issue #422, opened 2019-11-06
Restrict the color picker to a limited pool of color


EpokK:
Review the `ColorPicker` component to simplify it with a pool of colors. Right now, we are using a library for it, just remove it and revamp the component.

### Design
https://app.zeplin.io/project/5d3afcab62259b6cb60d7059/screen/5dc5773ae1948c239cc62dd7

### Note
- In the case the initial color doesn't exist in the pool, just display it as selected but don't add a new color in the dropdown.

Reported by Aude

gledunProton:
Reuse the same picker as folder component in ProtonMail

PR #430, opened 2019-11-12
Replace the third party color picker by a predefined list of colors

Swiip:
Fix: https://github.com/ProtonMail/react-components/issues/422

Works pretty well, yet I have some design questions:
- Should I put the list of colors in `proton-shared/lib/constants`? Somewhere else ?
- Is the color formating logic (parsing with `tinycolor`, checking for validity...) is still needed now that we have simplified the inputs?
- Should we "name" the colors, pass them through i18n for accessibility? (current "Use #d83078" is pretty bad)
- Should we use some visual indication about the button nature of the color dots? Some halo of a kind...

Capture d’écran 2019-11-12 à 15 22 09


mmso:
> Should I put the list of colors in proton-shared/lib/constants? Somewhere else ?

Yeah it could be a good idea to put these in proton-shared, like this we could take colors from this list, independently of the color picker.

> Is the color formating logic (parsing with tinycolor, checking for validity...) is still needed now that we have simplified the inputs?

Hmm this is tricky. I'd say if a color is passed that does not exist in the list, nothing should be highlighted, and this value should be allowed to stay. Otherwise I think we can trust it since it's predefined.

> Should we "name" the colors, pass them through i18n for accessibility? (current "Use #d83078" is pretty bad)

This is a good idea I think.

> Should we use some visual indication about the button nature of the color dots? Some halo of a kind...

Do you mean if it's selected? For that I'd say yes.

Looks good to me. I'd like some feedback from @EpokK though before we proceed with this.

components/input/ColorPicker.js
```diff
@@ -1,33 +1,57 @@
...
+    const { anchorRef, isOpen, toggle, close } = usePopperAnchor();
+
+    const handleClick = (color) => () => onChange({ hex: color });
 
     return (
-        <div className="relative">
-            <DropdownButton onClick={handleClick} hasCaret={true} {...rest}>
+        <>
+            <DropdownButton buttonRef={anchorRef} isOpen={isOpen} onClick={toggle} hasCaret={true}>
                 <Icon className="flex-item-noshrink" name="circle" color={iconColor} />
```
nico3333fr: Maybe add some text here using `sr-only` class (like "Choose color" or "choose color (blue selected)", otherwise it will say "button".

Swiip:
Just discovered that we already have something similar for the label colors.

Capture d’écran 2019-11-13 à 09 04 48

Shouldn't we merge a bit of logic ? Like the color list or even just put the component in the dropdown?

At least, there is a visual indicator for the selected value and the hover that I can reuse.

mmso:
@Swiip I think it would make sense to merge these two together. However @EpokK is the main requester of this feature and I don't know exactly what he wanted.

mmso:
@Swiip I discussed with @gledunproton and we should reuse the same logic as the other color picker

Swiip:
Improve accessibility should be done but perhaps later as it will now implicate label selector too. (And there is several color names to translate)

PR #433, opened 2019-11-13
Feat #89 - range selection design

nico3333fr:

- added some modifiers to help styling range selection
- added data attribute to help styling range selection

Fixes #89

@mmso  🙏 : when you click on only one day without creating a range, there is no aria-pressed="true", which makes styles fail.
image

## Snapshot

image


nico3333fr:
@mmso I just need that `isIntervalBoundEnd` is defined and put on the same element if I click on only one element (it is set up if I set a range of two days and come back to 1 day, but not on first try).

PR #449, opened 2019-11-25
Throw payment error for DuckDuckGo

EpokK:
DuckDuckGo mobile browser is not supporting [window.open](https://www.w3schools.com/jsref/met_win_open.asp) method required for our payment flow. So we have to handle this case properly for `card` and `paypal` payment methods.

- For PayPal we just display an error panel in the view.
- For card, we display the same error in the `PaymentVerificationModal` after getting the `Token` if the `Status` is not directly chargeable.

Manager: @jovanjovanovski 
Test server: https://vpnsettings.protonmail.blue (blue API)

### DIFF
--- a/containers/payments/PaymentVerificationModal.js
+++ b/containers/payments/PaymentVerificationModal.js
@@ -1,4 +1,4 @@
-import React, { useState, useRef, useEffect } from 'react';
+import React, { useState, useRef } from 'react';
 import PropTypes from 'prop-types';
 import {
     FormModal,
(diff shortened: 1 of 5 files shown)


containers/payments/PaymentVerificationModal.js
```diff
@@ -84,6 +88,12 @@ const PaymentVerificationModal = ({
         }
     };
 
+    useEffect(() => {
```
mmso: Put this in the useState up there

```suggestion
const [step, setStep] = useState(() => isDuckDuckGo() ? STEPS. DUCKDUCKGO  : STEPS.REDIRECT);
```

PR #457, opened 2019-11-28
Improve Black Friday

EpokK:
Capture d’écran 2019-11-28 à 13 47 52

- New component added: `Countdown`

Fix https://gitlab.protontech.ch/ProtonVPN/web/protonvpn-settings/issues/51
Manager Delong
Test server https://vpnsettings.protonmail.blue (`blue` API)

### DIFF
--- a/components/countdown/Countdown.js
+++ b/components/countdown/Countdown.js
@@ -1,0 +1,54 @@
+import React, { useEffect, useState } from 'react';
+import PropTypes from 'prop-types';
+import { c } from 'ttag';
+import { DAY, HOUR, MINUTE, SECOND } from 'proton-shared/lib/constants';
+import { isBefore, isAfter, differenceInMilliseconds } from 'date-fns';
+
+const EVERY_SECOND = SECOND;
+
+const Countdown = ({ start, end, separator = ' | ' }) => {
+    const [now, setNow] = useState(new Date());
+
+    useEffect(() => {
+        const intervalID = setInterval(() => {
+            setNow(new Date());
+        }, EVERY_SECOND);
+
+        return () => {
+            clearInterval(intervalID);
+        };
+    }, []);
+
+    if (start && isBefore(now, start)) {
+        return null;
+    }
+
+    if (isAfter(now, end)) {
+        return c('Countdown status').t`Expired`;
+    }
+
+    const diff = differenceInMilliseconds(end, now);
+    const days = Math.floor(diff / DAY);
+    const hours = Math.floor((diff % DAY) / HOUR);
+    const minutes = Math.floor((diff % HOUR) / MINUTE);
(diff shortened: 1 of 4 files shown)


components/countdown/Countdown.js
```diff
@@ -0,0 +1,54 @@
...
+    const minutes = Math.floor((diff % HOUR) / MINUTE);
+    const seconds = Math.floor((diff % MINUTE) / SECOND);
+
+    return (
+        <>
+            {[
+                c('Countdown unit').t`${days} days`,
+                c('Countdown unit').t`${hours} hours`,
+                c('Countdown unit').t`${minutes} minutes`,
+                c('Countdown unit').t`${seconds} seconds`
```
econdepe: Maybe we should take into account plurals here?

containers/payments/subscription/BlackFridayModal.js
```diff
@@ -32,19 +40,19 @@ const BlackFridayModal = ({ bundles = [], onSelect, ...rest }) => {
 
     const BILLED_DESCRIPTION = ({ cycle, amount, notice }) =>
         ({
-            [MONTHLY]: c('Title').jt`Billed as ${amount} (${notice})`,
-            [YEARLY]: c('Title').jt`Billed as ${amount} (${notice})`,
-            [TWO_YEARS]: c('Title').jt`Billed as ${amount} (${notice})`
+            [MONTHLY]: c('Title').jt`Billed as ${amount} ${NOTICES[notice]} for 1 month`,
```
econdepe: I would have the asterisks by the end: `Billed as ${amount} for 1 month ${NOTICES[notice]}`. I think they are also at the end in the proposal

econdepe:
Mmhh, what I said above is probably not a good idea given the async nature of the setState operations. Probably what you wrote is more stable

components/countdown/Countdown.js
```diff
@@ -0,0 +1,54 @@
...
+    const hours = Math.floor((diff % DAY) / HOUR);
+    const minutes = Math.floor((diff % HOUR) / MINUTE);
+    const seconds = Math.floor((diff % MINUTE) / SECOND);
+
+    return (
+        <>
+            {[
+                c('Countdown unit').ngettext(msgid`${days} day`, `${days} days`, days),
+                c('Countdown unit').ngettext(msgid`${hours} hour`, `${hours} hours`, days),
+                c('Countdown unit').ngettext(msgid`${minutes} minute`, `${minutes} minutes`, days),
```
econdepe: When `day` reaches 0, do we want to still display it? Same question for hours, when `hour` and `day` reach 0, and for minutes, when `day`, 'hour` and `minutes` reach 0

EpokK: Yes we want

PR #463, opened 2019-11-29
Add clear search buttons on search bars

Swiip:
Fix https://github.com/ProtonMail/proton-contacts/issues/275

Add clear buttons on both mobile and desktop search bars.

Capture d’écran 2019-11-29 à 17 16 46
Capture d’écran 2019-11-29 à 17 17 02

On desktop, the cross is hidden when the search field is empty or if the advanced feature is used.

### DIFF
--- a/containers/search/SearchDropdown.js
+++ b/containers/search/SearchDropdown.js
@@ -1,6 +1,14 @@
 import React, { useState, useEffect } from 'react';
 import PropTypes from 'prop-types';
-import { SearchInput, DropdownButton, Dropdown, Icon, usePopperAnchor, PrimaryButton } from 'react-components';
+import {
+    SearchInput,
+    DropdownButton,
+    Dropdown,
+    Icon,
+    usePopperAnchor,
+    PrimaryButton,
+    ResetButton
+} from 'react-components';
 import { c } from 'ttag';
 
 import { generateUID } from '../../helpers/component';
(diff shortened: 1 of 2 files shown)


containers/search/SearchDropdown.js
```diff
@@ -24,6 +32,12 @@ const SearchDropdown = ({
         close();
     };
 
+    const handleReset = () => {
+        updateSearch('');
```
econdepe: not sure if we should guard against the default reset action here, like we do for `handleSubmit`, since `ResetButton` is of type "reset"

containers/search/Searchbox.js
```diff
@@ -41,6 +47,12 @@ const Searchbox = ({ className = '', advanced, placeholder = '', value = '', onS
                 <Icon name="search" className="fill-white mauto searchbox-search-button-icon" />
                 <span className="sr-only">{c('Action').t`Search`}</span>
             </button>
+            {!advanced && search !== '' && (
```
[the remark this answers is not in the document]
Swiip: In the long run yes, but there is no app using the advanced mode yet (that I know off) and it requires changes to the CSS to have 2 icons in the right.

containers/search/SearchDropdown.js
Swiip: I asked myself too. But after some tests, I seen any problem.

containers/search/Searchbox.js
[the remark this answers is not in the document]
Swiip: I left a comment to be sure.

containers/search/SearchDropdown.js
Swiip: You made me doubt. I added the prevent default just to be sure.

PR #476, opened 2019-12-04
Fix - support drop down icon color

nico3333fr:

Icon of support drop down was not visible. 
image

- added `fill-currentColor` class (the icon will be the same color as the text in the button)

## Snapshot

image

image


nico3333fr:
Checked also on Signup, ok 
image

Issue #459, opened 2019-11-28
Storage pop over - update for visionary / lifetime accounts


matstaz:
When a user has a lifetime or visionary account. there is no way to upgrade to more storage or a higher tiered plan.

- [ ] remove the `UPGRADE` button if user has a visionary account
- [ ] remove the `UPGRADE` button if user has a lifetime account

PR #479, opened 2019-12-05
Do not show upgrade storage button when on visionary or lifetime plan

Yiin:
Fixes https://github.com/ProtonMail/react-components/issues/459

### DIFF
--- a/containers/sidebar/StorageSpaceStatus.js
+++ b/containers/sidebar/StorageSpaceStatus.js
@@ -5,28 +5,23 @@ import {
     CircularProgress,
     Dropdown,
     Icon,
+    Loader,
     generateUID,
     useUser,
     usePopperAnchor,
     useSubscription
 } from 'react-components';
 import humanSize from 'proton-shared/lib/helpers/humanSize';
-import { getPlanName, hasLifetime } from 'proton-shared/lib/helpers/subscription';
-import { PLANS } from 'proton-shared/lib/constants';
+import { hasVisionary, hasLifetime } from 'proton-shared/lib/helpers/subscription';
 
 import { classnames } from '../../helpers/component';
 
-const { VISIONARY } = PLANS;
-
 const StorageSpaceStatus = ({ upgradeButton }) => {
     const [{ MaxSpace, UsedSpace }] = useUser();
-    const [subscription = {}] = useSubscription();
+    const [subscription, loadingSubscription] = useSubscription();
     const [uid] = useState(generateUID('dropdown'));
     const { anchorRef, isOpen, toggle, close } = usePopperAnchor();
-    const planName = getPlanName(subscription);
-    const hasVisionaryPlan = planName === VISIONARY;
-    const hasLifetimePlan = hasLifetime(subscription);
-    const canUpgradeStorage = !hasVisionaryPlan && !hasLifetimePlan;
+    const canUpgradeStorage = !hasVisionary(subscription) && !hasLifetime(subscription);
 
     // round with 0.01 precision
     const usedPercent = Math.round((UsedSpace / MaxSpace) * 10000) / 100;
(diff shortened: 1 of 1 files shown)


containers/sidebar/StorageSpaceStatus.js
```diff
@@ -1,15 +1,32 @@
...
-const StorageSpaceStatus = ({ children }) => {
+const { VISIONARY } = PLANS;
+
+const StorageSpaceStatus = ({ upgradeButton }) => {
     const [{ MaxSpace, UsedSpace }] = useUser();
+    const [subscription = {}] = useSubscription();
     const [uid] = useState(generateUID('dropdown'));
     const { anchorRef, isOpen, toggle, close } = usePopperAnchor();
+    const planName = getPlanName(subscription);
+    const hasVisionaryPlan = planName === VISIONARY;
```
EpokK: In `proton-shared/lib/helpers/subscription` we already have a helper `hasVisionary`, just use it.

containers/sidebar/StorageSpaceStatus.js
```diff
@@ -1,15 +1,32 @@
...
+import { PLANS } from 'proton-shared/lib/constants';
 
 import { classnames } from '../../helpers/component';
 
-const StorageSpaceStatus = ({ children }) => {
+const { VISIONARY } = PLANS;
+
+const StorageSpaceStatus = ({ upgradeButton }) => {
     const [{ MaxSpace, UsedSpace }] = useUser();
+    const [subscription = {}] = useSubscription();
```
EpokK: `useSubscription` hook returns an `Array` with 2 parameters: `subscription` and `loadingSubscription`. You need to handle the loading state in case the subscription is not yet loaded in our models. 

Example:

``` jsx
const [subscription, loadingSubscription] = useSubscription();

if (loadingSubscription) {
   return <Loader />;
}

return <>Blablabla</>
```

containers/sidebar/StorageSpaceStatus.js
```diff
@@ -79,7 +96,7 @@ const StorageSpaceStatus = ({ children }) => {
 };
 
 StorageSpaceStatus.propTypes = {
-    children: PropTypes.node
+    upgradeButton: PropTypes.node
```
EpokK: Should be required, right?
```suggestion
    upgradeButton: PropTypes.node.isRequired
```

containers/sidebar/StorageSpaceStatus.js
```diff
@@ -1,15 +1,26 @@
...
+} from 'react-components';
 import humanSize from 'proton-shared/lib/helpers/humanSize';
+import { hasVisionary, hasLifetime } from 'proton-shared/lib/helpers/subscription';
 
 import { classnames } from '../../helpers/component';
 
-const StorageSpaceStatus = ({ children }) => {
+const StorageSpaceStatus = ({ upgradeButton }) => {
     const [{ MaxSpace, UsedSpace }] = useUser();
+    const [subscription = {}, loadingSubscription] = useSubscription();
```
EpokK: I don't think it's necessary to cast `subscription` to an `Object`.
```suggestion
    const [subscription, loadingSubscription] = useSubscription();
```

containers/sidebar/StorageSpaceStatus.js
```diff
@@ -1,15 +1,26 @@
...
     const [uid] = useState(generateUID('dropdown'));
     const { anchorRef, isOpen, toggle, close } = usePopperAnchor();
+    const canUpgradeStorage = !loadingSubscription && !hasVisionary(subscription) && !hasLifetime(subscription);
```
EpokK: Why not handling `loadingSubscription` with a `<Loader />` component? Let me know if it's not clear.

Yiin: I wasn't sure if rendering <Loader /> component in the place of the button is good idea, but I guess it's better than not showing the button at all.

PR #508, opened 2019-12-26
Proposal for centering months in mini-calendar

econdepe:
When displaying 6 weeks in the mini-calendar, center the month in it


mmso:
Seems fine to me. We have to think about doing the same thing in the monthly view too

PR #510, opened 2019-12-26
Offline notification

mmso:
Add offline notification and app version bad


containers/api/ApiProvider.js
```diff
@@ -25,29 +28,66 @@ const getSilenced = ({ silence } = {}, code) => {
...
                     : c('Error').t`No internet connection found`;
-                const isSilenced = getSilenced(e.config, code);
-                !isSilenced && createNotification({ type: 'error', text });
+                */
+                const id = createNotification({
+                    type: 'warning',
+                    text: (
+                        <OfflineNotification
+                            onRetry={() => {
+                                // TODO: What route when not logged in?
```
EpokK: Do you want to do something there?

PR #519, opened 2020-01-13
Review sidebar to include sub-sidebar

EpokK:
Capture d’écran 2020-01-13 à 18 07 26

### Note

* `SubSidebar` component removed.

### Related projects

* ProtonMail Settings: https://github.com/ProtonMail/proton-mail-settings/pull/279
* ProtonVPN Settings: https://github.com/ProtonMail/proton-vpn-settings/pull/330


EpokK:
> Can't you set default values to aria attributes depending on other props? Like ariaCurrent and isActive?

Not really, `aria-current` need to be defined manually if the value need to be different than `page` defined by `isActive` from `NavLink`: https://github.com/ReactTraining/react-router/blob/master/packages/react-router-dom/docs/api/NavLink.md#aria-current-string

PR #527, opened 2020-01-17
Allow React node for FormModal title

Swiip:
I need an Icon in the title. It works but type checking complains.


components/modal/FormModal.js
```diff
@@ -98,7 +98,7 @@ Modal.propTypes = {
     modalTitleID: PropTypes.string,
     onClose: PropTypes.func.isRequired,
     onSubmit: PropTypes.func,
-    title: PropTypes.string.isRequired,
+    title: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),
```
Yiin: `title: PropTypes.node,`
as `string` works as a node itself. (https://github.com/facebook/prop-types/blob/master/factoryWithTypeCheckers.js#L474)

EpokK:
```suggestion
    title: PropTypes.node,
```

Swiip:
Sorry for the delay. You're totaly right, node includes string. I duplicated the good practice to submit and close label. I also removed the required as no title is a valide case.

PR #528, opened 2020-01-17
Lazy load css for quill

econdepe:
The CSS for the quill editor was like 1/4 of our total css. It does not make sense to load it when loading the app, but rather "on demand" (i.e. when some component needs to use it)

### DIFF
--- a/components/input/RichTextEditor.js
+++ b/components/input/RichTextEditor.js
@@ -2,6 +2,9 @@ import React from 'react';
 import ReactQuill from 'react-quill';
 import Quill from 'quill';
 
+// This style will be lazy loaded by LazyRichTextEditor
+import 'react-quill/dist/quill.snow.css';
+
 const Block = Quill.import('blots/block');
 Block.tagName = 'div';
 Quill.register(Block);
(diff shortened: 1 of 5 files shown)


components/input/RichTextEditor.js
```diff
@@ -2,6 +2,9 @@ import React from 'react';
 import ReactQuill from 'react-quill';
 import Quill from 'quill';
 
+// This style will be lazy loaded by LazyRichTextEditor
```
nico3333fr: Quick question: 

there were 
```
@import '~react-quill/dist/quill.core.css';
@import '~react-quill/dist/quill.bubble.css';
@import '~react-quill/dist/quill.snow.css';
```
in react-components/index.scss, and now there is only `react-quill/dist/quill.snow.css` ?

Normal ?

econdepe: Yeah, core was just contained in snow, and bubble is a different theme that, as far as I could see, we were not using

nico3333fr: Fine, even lighter :)

PR #540, opened 2020-01-24
Feat - url for dark version of placeholder images

nico3333fr:

Missing support of dark mode version for placeholder illustrations.

- added `urlDarkVersion` property (not mandatory, in some cases, the image could fit in both cases ^^) 
- if specified, will be displayed
- added a class `hide-on-darkmode` in case of dark version (to hide the "light" version)

## Snapshot

image

### DIFF
--- a/containers/illustration/IllustrationPlaceholder.js
+++ b/containers/illustration/IllustrationPlaceholder.js
@@ -2,12 +2,14 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { classnames } from '../../helpers/component';
 
-const IllustrationPlaceholder = ({ className, title, text, url, uppercase, children }) => {
+const IllustrationPlaceholder = ({ className, title, text, url, urlDarkVersion = '', uppercase, children }) => {
     const info = typeof text === 'string' ? <p>{text}</p> : text;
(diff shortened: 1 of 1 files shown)


nico3333fr:
Yes, we already had this discussion with @econdepe, this rely on CSS.

econdepe:
We introduced a dummy class in <body> in the PR https://github.com/ProtonMail/react-components/pull/537 for an unrelated issue (Nico needed it in the subscription flow). Since the class is there, it does make sense to re-use it as in this PR, but I agree a React hook would look nicer.

mmso:
I think this suggestion looks slightly better than this PR though:

```js
// somewhere
export const getDarkOrLight = (light, dark) => {
    return document.body.classList.contains('isDarkMode') ? dark : light;
};
// elsewhere
<IllustrationPlaceholder
    url={getDarkOrLight(lighImageImport, darkImageImport)}
    ...
/>
```

I think it's as fast as doing this, and then we can refactor with the ThemeProvider later

econdepe:
Sorry about the force-pushes (the first one was unintentional...). The helper is available in https://github.com/ProtonMail/proton-shared/pull/114
This PR now just moves a constant to proton-shared

PR #546, opened 2020-01-30
Fix - storage popover margins

nico3333fr:

The margin on Close button on storage popover was not the same as for other popovers/modals

- set it up to exact same values as for modals/popovers
- bonus, added a text for better a11y  (empty buttons are badly vocalized)

## Snapshot

image

### DIFF
--- a/containers/sidebar/StorageSpaceStatus.js
+++ b/containers/sidebar/StorageSpaceStatus.js
@@ -57,9 +57,10 @@ const StorageSpaceStatus = ({ upgradeButton }) => {
                 size="auto"
             >
                 <div className="dropDown-content">
-                    <div className="absolute top-right mt0-5 mr0-5">
-                        <button className="flex flex-items-center">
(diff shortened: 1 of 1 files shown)


containers/sidebar/StorageSpaceStatus.js
```diff
@@ -57,9 +57,10 @@ const StorageSpaceStatus = ({ upgradeButton }) => {
                 size="auto"
             >
                 <div className="dropDown-content">
-                    <div className="absolute top-right mt0-5 mr0-5">
-                        <button className="flex flex-items-center">
+                    <div className="absolute top-right mt0-5r mr0-5r">
+                        <button className="flex flex-items-center" title={c('Action').t`Close`}>
```
EpokK: It's important to always define `type="button"` if you are not using `<Button>` component. A `button` in a form is considered as a submit button by default.

```suggestion
                        <button type="button" className="flex flex-items-center" title={c('Action').t`Close`}>
```

